PR #9, opened 2022-01-16
Washer-Dryer information enhancements

valorp:
Some appliance.py enhancements:
* attr_value_to_int: added method to get int from attribute (JSON API response string)
* attr_value_secs_to_mins_int: get an int for mins from seconds
* load_from_file: needs to be moved out of this class, but allows a hard-coded (for now) JSON payload to be specified instead of API call for testing purposes.

washerdryer.py enhancements:
* added "pretty" outputs for machine state, started on "pretty" cycle status responses
* Added cycle time remaining attributes

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -2,6 +2,7 @@ import aioconsole
 import argparse
 import asyncio
 import logging
+import math
 from cli_ac_menu import show_aircon_menu
 from cli_washerdryer_menu import show_washerdryer_menu
 
@@ -48,8 +49,27 @@ async def start():
         return
 
     if args.list:
-        print(appliance_manager.aircons)
-        print(appliance_manager.washer_dryers)
+        for index, cooler in enumerate(appliance_manager.aircons):
+            print(index, cooler)
+        for index, hole in enumerate(appliance_manager.washer_dryers):
+            print(index, hole)
+        applianceelementinput = input("Enter 0-" + str(index) + " for appliance details, otherwise type 'n': ")
+        try:
+            applianceelement = int(applianceelementinput)
+        except ValueError:
+            return
+        print("Appliance selected: " + str(applianceelement))
+
+        if hasattr(appliance_manager, "aircons"):
+            if applianceelement < len(appliance_manager.aircons):
+                args.said = appliance_manager.aircons[applianceelement]["SAID"]
+                await show_aircon_menu(backend_selector, auth, args.said)
+                return
+        if hasattr(appliance_manager, "washer_dryers"):
+            if applianceelement < len(appliance_manager.washer_dryers[applianceelement]["SAID"]):
+                args.said = appliance_manager.washer_dryers[applianceelement]["SAID"]
+                await show_washerdryer_menu(backend_selector, auth, args.said)
+                return
         return
 
     if not args.said:
--- a/cli_washerdryer_menu.py
+++ b/cli_washerdryer_menu.py
@@ -8,6 +8,7 @@ async def show_washerdryer_menu(backend_selector, auth, said):
         print("\n")
         print(30 * "-", "MENU", 30 * "-")
         print("u. Update status from server")
+        print("l. Load response from file")
         print("p. Print status")
         print("v. Print raw status")
         print("c. Custom command")
@@ -15,14 +16,20 @@ async def show_washerdryer_menu(backend_selector, auth, said):
         print(67 * "-")
 
     def print_status(wd: WasherDryer):
-        print("online: " + str(wd.get_online()))
-        print("state: " + str(wd.get_machine_state()))
+        print("Online: " + str(wd.get_online()))
+        print("State: " + str(wd.get_machine_state()))
+        print("Pretty state: " + str(wd.get_machine_state_pretty()))
+        print("Time Remaining (mins): " + str(wd.get_time_remaining_mins()))
+        print("Total time remaining (delay + cycle): " + str(wd.get_total_delay_and_cycle_time_remaining_mins_pretty()))
+
         print("sensing: " + str(wd.get_cycle_status_sensing()))
         print("filling: " + str(wd.get_cycle_status_filling()))
         print("soaking: " + str(wd.get_cycle_status_soaking()))
         print("washing: " + str(wd.get_cycle_status_washing()))
         print("rinsing: " + str(wd.get_cycle_status_rinsing()))
         print("spinning: " + str(wd.get_cycle_status_spinning()))
+        wd.get_cycle_status_pretty()
+        print("Cylce status: " + str(wd.get_cycle_status_pretty()))
 
     def attr_upd():
         print("Attributes updated")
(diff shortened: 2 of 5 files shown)


cli.py
```diff
@@ -48,8 +49,27 @@ def attr_upd():
...
     if args.list:
-        print(appliance_manager.aircons)
-        print(appliance_manager.washer_dryers)
+        for index, cooler in enumerate(appliance_manager.aircons):
+            print(index, cooler)
+        for index, hole in enumerate(appliance_manager.washer_dryers):
+            print(index, hole)
+        applianceelementinput = input("Enter 0-" + str(index) + " for appliance details, otherwise type 'n': ")
+        try:
+            applianceelement = int(applianceelementinput)
```
abmantis, line 58:
```suggestion
        appliance_input = input("Enter 0-" + str(index) + " for appliance details, otherwise type 'n': ")
        try:
            selected_appliance = int(appliance_input)
```

cli.py
```diff
@@ -48,8 +49,27 @@ def attr_upd():
...
+        except ValueError:
+            return
+        print("Appliance selected: " + str(applianceelement))
+
+        if hasattr(appliance_manager, "aircons"):
```
abmantis, line 63: someone could have both an aircon an dryer. 
but we can improve this in the future, so I'm fine with keeping this as it is for now :)

cli.py
abmantis: we should use lower_case_with_underscores for names pls

cli_washerdryer_menu.py
```diff
@@ -8,21 +8,28 @@ def print_menu():
...
+        print("Time Remaining (mins): " + str(wd.get_time_remaining_mins()))
+        print("Total time remaining (delay + cycle): " + str(wd.get_total_delay_and_cycle_time_remaining_mins_pretty()))
+
         print("sensing: " + str(wd.get_cycle_status_sensing()))
         print("filling: " + str(wd.get_cycle_status_filling()))
         print("soaking: " + str(wd.get_cycle_status_soaking()))
         print("washing: " + str(wd.get_cycle_status_washing()))
         print("rinsing: " + str(wd.get_cycle_status_rinsing()))
         print("spinning: " + str(wd.get_cycle_status_spinning()))
+        wd.get_cycle_status_pretty()
```
abmantis, line 31:
```suggestion
```

cli_washerdryer_menu.py
```diff
@@ -8,21 +8,28 @@ def print_menu():
         print("\n")
         print(30 * "-", "MENU", 30 * "-")
         print("u. Update status from server")
+        print("l. Load response from file")
```
abmantis, line 11:
```suggestion
        print("l. Load response from file (for testing purposes)")
```

whirlpool/appliance.py
```diff
@@ -68,6 +69,11 @@ def _set_attribute(self, attribute, value, timestamp):
     def said(self):
         return self._said
 
+    def load_from_file (self):
+        with open("APISamples/Maytag - Dryer Sensing-Wet, Bulky Items cycle, Temp 4, Dryness 3, Wrinkle prevent low, extra power.json") as file:
```
abmantis, line 73:
```suggestion
        with open("APISamples/mock_appliance_data.json") as file:
```
Lets make this a bit more generic. You could then just use a symbolic link named `mock_appliance_data.json` to choose which json to load.

whirlpool/appliance.py
```diff
@@ -68,6 +69,11 @@ def _set_attribute(self, attribute, value, timestamp):
...
+            self._data_dict = json.load(file)
+        return True
```
abmantis, line 75:
```suggestion
```

whirlpool/appliance.py
```diff
@@ -122,6 +128,21 @@ def bool_to_attr_value(self, b: bool):
     def attr_value_to_bool(self, val: str):
         return None if val is None else val == SETVAL_VALUE_ON
 
+    def attr_value_to_int(self, val: str):
+        value = 0
+        try:
+            value = int(val)
+        except ValueError:
+            return 0
```
abmantis, line 136: I think we should return None if it is not an in, otherwise we will have no way to distinguish an invalid value from a valid zero one upper in the lib or HA.

whirlpool/appliance.py
```diff
@@ -122,6 +128,21 @@ def bool_to_attr_value(self, b: bool):
...
+        return value
+
+    def attr_value_secs_to_mins_int(self, val: str):
+        val_int = self.attr_value_to_int(val)
+        mins = math.floor(val_int/60)
+        
+        return mins
+
+
```
abmantis, line 145:
```suggestion
```
The lib should just return secs. HA will take care of converting to minutes if it makes sense there.

whirlpool/washerdryer.py
```diff
@@ -79,6 +81,59 @@ class MachineState(Enum):
     MachineState.SystemInit: ATTRVAL_MACHINE_STATE_SYSTEMINIT,
 }
 
+MACHINE_STATE_MAP_PRETTY = {
```
abmantis, line 84: The lib should not have "pretty" strings, but only Enums, because it is just a data provider and pretty strings are not the best way to pass data between code components. HA should take care of prettifying, since HA is the one showing the UI to the user.
If you don't mind, I ask you to please do remove the "pretty" things :sweat_smile:

whirlpool/washerdryer.py
```diff
@@ -91,6 +146,51 @@ def get_machine_state(self):
...
+            print (machine_state_raw, ATTRVAL_MACHINE_STATE_RUNNINGPOSTCYCLE)
+            if self.get_time_remaining_mins() == 0:
+                return MACHINE_STATE_MAP_PRETTY[MachineState.Complete]
+        return machine_state_pretty
+
+
+    def get_time_remaining_mins(self):
+        return self.attr_value_secs_to_mins_int(self.get_attribute(ATTR_CYCLE_TIME_REMAINING))
+
+    #Handles delay time
```
abmantis, line 169:
```suggestion
```
function name is already explicit

whirlpool/washerdryer.py
```diff
@@ -91,6 +146,51 @@ def get_machine_state(self):
...
+    def get_delay_time_remaining_mins(self):
+        try: 
+            return self.attr_value_secs_to_mins_int(self.get_attribute(ATTR_DELAY_TIME_REMAINING))
+        except:
+            return 0
+
+    def get_total_delay_and_cycle_time_remaining_mins(self):
+        delay_time_remaining = self.get_delay_time_remaining_mins()
+        cycle_time_remaining = self.get_time_remaining_mins()
+        return delay_time_remaining + cycle_time_remaining
```
abmantis, line 179:
```suggestion
```
The sum would also be done in an upper layer like HA. This washerdryer class aims to provide direct values for most things, only converting types (int, bool, float) and enumerations to python enums, but it should not have any result manipulation.

whirlpool/washerdryer.py
```diff
@@ -79,6 +81,59 @@ class MachineState(Enum):
...
+    MachineState.SystemInit: "Initializing",
+}
+
+class WashCycleState (Enum):
+    Sensing = False
+    Filling = False
+    Soaking = False
+    Washing = False
+    Rinsing = False
+    Spinning= False
```
abmantis, line 112: we should not have an enum class with all the values being equal (False). I'm not sure about the approach done here regarding those values.

abmantis (changes requested):

PR #13, opened 2022-08-29
Renew auth on eventsocket and improve reconnect handling

mkmer:
I believe this will address the authkey renewal.
Also found that connection timeout needs to be set or the event socket hangs at connect when the host decides not to connect...


whirlpool/washerdryer.py
```diff
@@ -108,3 +108,6 @@ def get_cycle_status_rinsing(self):
 
     def get_cycle_status_spinning(self):
         return self.attr_value_to_bool(self.get_attribute(ATTR_CYCLE_STATUS_SPINNING))
+
+    def get_all_attributes(self):
+        return self._data_dict.get("attributes")
```
mkmer, line 113: Sorry, this isn't needed for the auth fix.  It's an artifact of some testing I was doing.

abmantis:
I noticed that this is getting a few extra changes in addition to the websocket fixes, right? 
If so, please keep this PR only for the websocket thing. You can open another one for the rest. 

I've made the PR a Draft. Please mark it as "Ready" when you feel it is ready for review (after removing non-websocket changes).

mkmer:
I could do that... or we can have on PR that will fit with the HA release.
HA with the laundry appliances will need everything I have added.  Will
that make sense?

abmantis:
The release is done from the master branch, so the PR needs to get merged first. There is no problem in having things entering master in short PRs :) 
It is usually easier to review and work on smaller PRs, that was why I was asking to have smaller and separate ones.

If you need  (for testing and so on) you can always base one PR branch on the other PR branch so you can have everything in the second one, for example.

whirlpool/appliance.py
```diff
@@ -86,12 +86,16 @@ async def fetch_data(self):
             return False
 
         uri = f"{self._backend_selector.base_url}/api/v1/appliance/{self._said}"
-        self._data_dict = None
+        #self._data_dict = None # Clearing the _dict causes HA to complain about missing attributes
```
abmantis, line 89: Lets leave this out of this PR, since it is not related to the websocket

whirlpool/backendselector.py
```diff
@@ -27,7 +27,6 @@ class Region(Enum):
     Region.US: {"base_url": "https://api.whrcloud.com"},
 }
 
-
```
abmantis, line 30: this was formatted with black, and we should keep it. please also try to use black while adding code to keep it formatted
I plan to add a github job for validation, but have not yet.

whirlpool/appliance.py
```diff
@@ -86,12 +86,16 @@ async def fetch_data(self):
             return False
 
         uri = f"{self._backend_selector.base_url}/api/v1/appliance/{self._said}"
-        self._data_dict = None
+        #self._data_dict = None # Clearing the _dict causes HA to complain about missing attributes 
         async with async_timeout.timeout(30):
-            async with self._session.get(uri) as r:
+            async with self._session.get(uri,headers=self._create_headers()) as r:
```
abmantis, line 91: the session already has the correct headers, so we do not need to re-set them on the get call.

whirlpool/eventsocket.py
```diff
@@ -14,20 +14,42 @@
...
+STATUS_TLS_HANDSHAKE_ERROR = 1015
+STATUS_UNAUTHORIZED = 3000
+
+RECONNECT_TIMEOUT = 30
+GOING_AWAY_TIMEOUT = 60*5 - RECONNECT_TIMEOUT
+
 
 class EventSocket:
-    def __init__(self, url, access_token, said, msg_listener: Callable[[str], None]):
+    def __init__(self, url, auth:Auth, said, msg_listener: Callable[[str], None],fetch_data:Callable):
```
abmantis, line 40:
```suggestion
    def __init__(self, url, auth: Auth, said, msg_listener: Callable[[str], None], conn_up_listener: Callable):
```
Lets call it conn_up_listener, to make it generic: from the Websocket PoV, it does not know about what the Appliance does, it simply exposes callbacks that the Appliance can act on.

conn_up_listener = connection up

whirlpool/eventsocket.py
```diff
@@ -14,20 +14,42 @@
...
         self._url = url
-        self._access_token = access_token
+        self._auth = auth
         self._said = said
         self._msg_listener = msg_listener
         self._running = False
         self._websocket: aiohttp.ClientWebSocketResponse = None
         self._run_future = None
+        self._fetch_data = fetch_data
```
abmantis, line 48:
```suggestion
        self._conn_up_listener = conn_up_listener
```

whirlpool/eventsocket.py
```diff
@@ -14,20 +14,42 @@
 
 RECV_MSG_MATCHER = re.compile("{(.*)}\\x00")
 
+# closing frame status codes for websocket.
```
abmantis, line 17:
```suggestion
```
Instead of the comment, prefix the codes with WS_, for example WS_STATUS_GOING_AWAY.

whirlpool/eventsocket.py
```diff
@@ -14,20 +14,42 @@
...
+STATUS_NORMAL = 1000
```
abmantis, line 18: please keep only the ones that are currently used

whirlpool/eventsocket.py
```diff
@@ -14,20 +14,42 @@
...
+STATUS_MESSAGE_TOO_BIG = 1009
+STATUS_INVALID_EXTENSION = 1010
+STATUS_UNEXPECTED_CONDITION = 1011
+STATUS_SERVICE_RESTART = 1012
+STATUS_TRY_AGAIN_LATER = 1013
+STATUS_BAD_GATEWAY = 1014
+STATUS_TLS_HANDSHAKE_ERROR = 1015
+STATUS_UNAUTHORIZED = 3000
+
+RECONNECT_TIMEOUT = 30
```
abmantis, line 35:
```suggestion
RECONNECT_DELAY = 30
```
It seems like a delay to reconnect, and not a connection timeout, right?

whirlpool/eventsocket.py
```diff
@@ -14,20 +14,42 @@
...
+GOING_AWAY_TIMEOUT = 60*5 - RECONNECT_TIMEOUT
```
abmantis, line 36:
```suggestion
GOING_AWAY_DELAY = 60*5 - RECONNECT_DELAY
```

whirlpool/eventsocket.py
```diff
@@ -45,12 +67,12 @@ async def _recv_msg(self, websocket: aiohttp.ClientWebSocketResponse):
     async def _run(self):
         if not self._running:
             return
-
-        timeout = aiohttp.ClientTimeout(total=None)
+        await self._fetch_data()
+        timeout = aiohttp.ClientTimeout(total=None,sock_connect=60)
```
abmantis, line 71:
```suggestion
        timeout = aiohttp.ClientTimeout(total=None, sock_connect=60)
```

whirlpool/eventsocket.py
```diff
@@ -45,12 +67,12 @@ async def _recv_msg(self, websocket: aiohttp.ClientWebSocketResponse):
     async def _run(self):
         if not self._running:
             return
-
-        timeout = aiohttp.ClientTimeout(total=None)
+        await self._fetch_data()
```
abmantis, line 70: put this after after connection is done and subscribed, since it is only there when the data starts coming from the WS

whirlpool/eventsocket.py
```diff
@@ -72,7 +94,17 @@ async def _run(self):
                         LOGGER.debug(
                             f"Stopping receiving. Message type: {str(msg.type)}"
                         )
+
+                        if not self._auth.is_access_token_valid() or msg.data==STATUS_UNAUTHORIZED: 
+                            LOGGER.debug("auth key expired, doing reauth now") 
+                            await self._auth.do_auth()
+
+                        elif msg.data==STATUS_GOING_AWAY: # if going away is issued without and expired key, it appears the server is going down.
+                            LOGGER.debug(f"Received Going Away Message 1001: Waiting {GOING_AWAY_TIMEOUT} seconds")
```
abmantis, line 103:
```suggestion
                            LOGGER.warning(f"Received Going Away message: Waiting {GOING_AWAY_TIMEOUT} seconds")
```

whirlpool/eventsocket.py
```diff
@@ -72,7 +94,17 @@ async def _run(self):
                         LOGGER.debug(
                             f"Stopping receiving. Message type: {str(msg.type)}"
                         )
+
+                        if not self._auth.is_access_token_valid() or msg.data==STATUS_UNAUTHORIZED: 
+                            LOGGER.debug("auth key expired, doing reauth now") 
+                            await self._auth.do_auth()
+
+                        elif msg.data==STATUS_GOING_AWAY: # if going away is issued without and expired key, it appears the server is going down.
```
abmantis, line 102:
```suggestion
                        elif msg.data==STATUS_GOING_AWAY:
```
No need for comment since it is part of the WS spec

whirlpool/eventsocket.py
```diff
@@ -72,7 +94,17 @@ async def _run(self):
...
+                            LOGGER.debug(f"Received Going Away Message 1001: Waiting {GOING_AWAY_TIMEOUT} seconds")
+                            await asyncio.sleep(GOING_AWAY_TIMEOUT) # be nice and let them reboot or whatever
```
abmantis, line 104:
```suggestion
                            # give the server some time to come back up
                            await asyncio.sleep(GOING_AWAY_TIMEOUT) 
```

whirlpool/eventsocket.py
```diff
@@ -84,10 +116,12 @@ async def _run(self):
 
             self._websocket = None
 
-        if self._running and self._reconnect_tries > 0:
```
abmantis, line 87: I don't think we should remove this.
We can retry 3 times fast, but then we should have a bigger wait time (1hour) after that.
If the server goes down, we don't want HA to start hammering it with a lot of requests while it is trying to boot up, and risk bringing it down again with a DOS :smile:

abmantis (changes requested):
Thanks for your work and for splitting the PRs!

whirlpool/appliance.py
mkmer: This is related to the websocket reconnect - part of the reconnect is to fetch the current data (via a fetch).  When fetch clears the dict and HA tries to poll the data it complains about missing keys and the integration dies.  Can't fix the eventsocket reconnect without this.

whirlpool/backendselector.py
mkmer: Not even sure why it was removed... sorry

whirlpool/appliance.py
mkmer: They actually are not up to date, and are holding the OLD auth key.  When the eventsocket renews the key and tries to restart, these headers are not utilizing the new auth key (and it fails to update).

whirlpool/eventsocket.py
mkmer: I was hoping to preserve the order (elif) with a comment as to why it's in this order... suppose we can put in the bit bucket.
Unfortunately, they are issuing a 1001 when the key expires, and when they are rebooting or whatever else. 
In this configuration, if we get a disconnect and the key is expired, just renew and try again.  If it's not expired, wait longer....

whirlpool/eventsocket.py
mkmer: Find a try at this - fast retry for 3 attempts then slow down forever....

mkmer: On this note - they do seem to send the 1001 message randomly - sometimes multiple times quickly, other times once / hour, and I have seen it go to auth key expiration (aka no 1001's for 2 or 3 days except for key renewal)   I'll get this change running in my local HA and watch for any odd behavior.

whirlpool/eventsocket.py
```diff
@@ -84,10 +119,23 @@ async def _run(self):
...
+        if self._running:
+
+            self._reconnect_tries -= 1
+            if self._reconnect_tries < 0:
+                self._reconnect_tries = 0
+                LOGGER.info(
+                    f"Waiting to reconnect long delay {RECONNECT_LONG_DELAY} seconds"
+                )
+                await asyncio.sleep(RECONNECT_LONG_DELAY)
+                # give server some time to come back up
```
abmantis, line 131:
```suggestion
                # give server some time to come back up
                await asyncio.sleep(RECONNECT_LONG_DELAY)
```

abmantis (changes requested):
just some minor changes now

whirlpool/appliance.py
mkmer: Seems to make sense - I can't test the post parts of the code, but taking the headers out of start_http_session doesn't appear to break anything...

whirlpool/eventsocket.py
```diff
@@ -72,7 +90,24 @@ async def _run(self):
...
+                            LOGGER.debug("auth key expired, doing reauth now")
+                            await self._auth.do_auth()
+
+                        elif msg.data == WS_STATUS_GOING_AWAY:
+                            LOGGER.debug(
+                                f"Received Going Away message: Waiting for {GOING_AWAY_DELAY} seconds"
+                            )
+                            await asyncio.sleep(
+                                GOING_AWAY_DELAY
+                            )  # be nice and let them reboot or whatever
```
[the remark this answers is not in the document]
mkmer: Sorry - thought I did that.

whirlpool/appliance.py
```diff
@@ -86,12 +86,15 @@ async def fetch_data(self):
...
         async with async_timeout.timeout(30):
-            async with self._session.get(uri) as r:
+            async with self._session.get(uri, headers=self._create_headers()) as r:
                 self._data_dict = json.loads(await r.text())
                 if r.status == 200:
                     return True
+                elif r.status == 401:
+                    await self._auth.do_auth()
+                    await self.start_http_session()
+
```
mkmer, line 97: Now that the headers are in PUT/GET maybe we don't need to restart http session here?

whirlpool/eventsocket.py
```diff
@@ -45,49 +61,95 @@ async def _recv_msg(self, websocket: aiohttp.ClientWebSocketResponse):
...
+
+                        match = RECV_MSG_MATCHER.findall(msg.data)
+                        if not match:
+                            continue
+                        self._msg_listener("{" + match[0] + "}")
+        except (
+            aiohttp.ClientConnectionError,
+            aiohttp.ClientError,
+            asyncio.TimeoutError,
+        ):
```
mkmer, line 130: This except grouping was found in multiple examples, but the aiohttp docs do not say it raises them.  I put this in due to a couple of instances where the eventsocket didn't restart and appeared to die with the "connecting to...." debug and nothing more happened - leading me to believe the connect or send/receive messages raised some error that stopped the thread.

whirlpool/eventsocket.py
```diff
@@ -61,63 +61,76 @@ async def _recv_msg(self, websocket: aiohttp.ClientWebSocketResponse):
...
+            LOGGER.info("Web socket could not connect")
```
abmantis, line 131:
```suggestion
            LOGGER.error("Websocket could not connect")
```

whirlpool/appliance.py
abmantis: yeah, we can remove that call

mkmer:
The socket has been alive for a week on my HA instance, happy to report recovery from multiple rauths and 1001 closes.  I'm feeling like it's solid :)

abmantis (approved):

abmantis:
Thanks for your work!

PR #26, opened 2022-11-10
Include KitchenAid brand

MizterB:
Enable communication with KitchenAid brand appliances

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -22,7 +22,7 @@ parser = argparse.ArgumentParser()
 parser.add_argument("-e", "--email", help="Email address")
 parser.add_argument("-p", "--password", help="Password")
 parser.add_argument(
-    "-b", "--brand", help="Brand (whirlpool/maytag)", default="whirlpool"
+    "-b", "--brand", help="Brand (whirlpool/maytag/kitchenaid)", default="whirlpool"
 )
 parser.add_argument("-r", "--region", help="Region (EU/US)", default="EU")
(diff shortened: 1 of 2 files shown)


abmantis (approved):
Thanks!

mkmer:
This is an interesting development - I wonder how they do this with mixed brands on a users account...  My account is brand maytag (in the US) even though the appliances are whirlpool.  
I think we will have some challenges on the HA integration configuration as we move forward :(

abmantis:
> This is an interesting development - I wonder how they do this with mixed brands on a users account... My account is brand maytag (in the US) even though the appliances are whirlpool. I think we will have some challenges on the HA integration configuration as we move forward :(

But you use the Whirlpool app, right? The Whirlpool app here (in EU) uses the whirlpool client_id. I assume that in the US it simply uses maytag for both Whirlpool + Maytag appliances.

mkmer:
Correct - I'm in the US, using whirlpool android app AND have whirlpool appliances, yet must set the maytag brand for a successful login.  
So far, we didn't put a brand option on the HA config flow, and at least as far as I know, it will be confusing in the US.  Now that we have a kitchenaid brand too (assuming it's needed in EU?) we will need to put the brand selector in the config flow, and it will more than likely be confusing.  I suppose we could address it with the docs.

abmantis:
In the Brand option, I think we could also have a default "Auto" brand, where the integration tries to auth one by one, and selects the one that succeeds.

PR #27, opened 2022-11-15
Add external session to API

mkmer:
HA is asyncIO timout during startup - found that we are not raising/re-raising the timeouts for HA to catch. 
Once these are raised, HA will try to connect again  by raising "ConfigEntryNotReady"


whirlpool/auth.py
```diff
@@ -95,8 +95,11 @@ async def _do_auth(self, refresh_token):
                         return json.loads(await r.text())
                     elif refresh_token:
                         return await self._do_auth(refresh_token=None)
+                    elif r.status != 200:
+                        raise aiohttp.ClientConnectionError
```
abmantis, line 99:
```suggestion
```

This is a real error, and we should not treat it as a timeout on HA. If there is an HTTP error coming from the server, HA should just show an error message, no retry.

whirlpool/auth.py
```diff
@@ -95,8 +95,11 @@ async def _do_auth(self, refresh_token):
...
                     return None
-
+        except (aiohttp.ClientConnectionError, asyncio.TimeoutError) as ex:
+            raise ConnectionError("Cannot connect")
```
abmantis, line 102:
```suggestion
        except asyncio.TimeoutError as ex:
            raise ex
```
Could we simply re-raise the timeout? Errors that are not timeouts should not be treated as such.

whirlpool/washerdryer.py
```diff
@@ -107,3 +107,6 @@ def get_cycle_status_rinsing(self):
 
     def get_cycle_status_spinning(self):
         return self.attr_value_to_bool(self.get_attribute(ATTR_CYCLE_STATUS_SPINNING))
+
+    def get_all_attributes(self):
+        return self._data_dict.get("attributes")
```
abmantis, line 112: does not seem related. oops? :)

abmantis (changes requested):

whirlpool/auth.py
mkmer: This error is one I added to my Aladdin Connect integration (copied from some other integration which I've forgotten at this point) - If the server complains with "not 200" should retry, say the server is busy, or some other complaint besides not authorized.  Otherwise the integration just hangs and the entities become "unavailable" and "no longer provided".
I suppose it could be some other locally defined error..

whirlpool/washerdryer.py
mkmer: Yes, I'm not sure how that got in here, had to do with some testing quite a bit ago. Sorry about that.

whirlpool/auth.py
mkmer: I think we should keep the ClientConnectionError and TimeoutErrors.... see proposed

mkmer:
Still not sure I like how we handle !=200 messages - but I suppose they end up with an authentication error in the integration. 

It appears I missed the rasie ConfigEntryAuthFailed in the integration - the asyncio timout would toss that, even though it's not "true"

whirlpool/auth.py
```diff
@@ -96,7 +96,8 @@ async def _do_auth(self, refresh_token):
                     elif refresh_token:
                         return await self._do_auth(refresh_token=None)
                     return None
-
+        except (aiohttp.ClientConnectionError, asyncio.TimeoutError) as ex:
+            raise ex
```
mkmer, line 100: The primary point is to let HA know there was a connect error or timeout so that it can try again later.  The original code path (after I found the error in the integration) will result in an invalid password declaration (which is not right).

abmantis: Maybe we could just do `async with aiohttp.ClientSession() as session:` after line 89, to get the session instance, and then we don't need the `finally` block at all. That means we can just remove the try/except/finally, and every exception will bubble up to HA.
What do you think?

abmantis (changes requested):

whirlpool/auth.py
mkmer: Looking over the requests the devs made for AladdinConnect, we are actually supposed to get  the session from HA with async_get_clientsession() and pass into the API, then use that session, and never close it in the API - HA takes care of it.

https://github.com/home-assistant/core/blob/8345176ccc1ca17e5b12960be13447e14fe23544/homeassistant/components/aladdin_connect/__init__.py#L27-L28

https://github.com/mkmer/AIOAladdinConnect/blob/e61607f56bbc9bedb83066c54e8f77cf4fb63e12/AIOAladdinConnect/__init__.py#L55

That would allow us to do the same, without finally, and the session would be passed in....
Thoughts?

abmantis: Yeah, that is a better approach!

mkmer: I think I got it setup with the "external" client session - appears to work with the cli tests.  Probably a mistake or two in there :(

mkmer:
I'm not sure how to fix the tests for passing in the ClientSession.

whirlpool/auth.py
```diff
@@ -96,9 +101,8 @@ async def _do_auth(self, refresh_token):
                     elif refresh_token:
                         return await self._do_auth(refresh_token=None)
                     return None
-
-        finally:
-            await session.close()
+        except (aiohttp.ClientConnectionError, asyncio.TimeoutError) as ex:
+            raise ex
```
abmantis, line 105: Since we just re-raise the exceptions, I think we can just remove the try/except, or not?

whirlpool/appliance.py
```diff
@@ -151,14 +152,16 @@ async def disconnect(self):
         await self.stop_event_listener()
 
     async def start_http_session(self):
-        await self.stop_http_session()
-        self._session = aiohttp.ClientSession()
+        # await self.stop_http_session()
```
abmantis, line 155: lets remove this method completely since it does not make sense anymore

whirlpool/appliance.py
```diff
@@ -151,14 +152,16 @@ async def disconnect(self):
...
+        # self._session = aiohttp.ClientSession()
+        return
 
     async def stop_http_session(self):
```
abmantis, line 159: also lets remove this one

tests/test_aircon.py
```diff
@@ -117,14 +120,20 @@ async def test_setters(
     appliance_http_client_mock: AiohttpClientMocker,
     backend_selector_mock: BackendSelectorMock,
     auth_mock: MagicMock,
+    http_client_mock: AiohttpClientMocker,
```
abmantis, line 123:
```suggestion
```
it is unused

tests/test_aircon.py
```diff
@@ -71,8 +72,10 @@ async def test_attributes(
     mock_appliance_http_get(
         appliance_http_client_mock, backend_selector_mock, SAID, DATA1
     )
-
-    aircon = Aircon(backend_selector_mock, auth_mock, SAID)
+    appliance_http_client_mock.create_session(asyncio.get_event_loop())
```
abmantis, line 75: since the session is created on the test now, at the end of the test we should close the session, otherwise we get "Unclosed client session" warnings.
Same for the other tests

cli.py
```diff
@@ -51,38 +52,46 @@ def attr_upd():
         return
 
     backend_selector = BackendSelector(selected_brand, selected_region)
+    session = aiohttp.ClientSession()
 
-    auth = Auth(backend_selector, args.email, args.password)
+    auth = Auth(backend_selector, args.email, args.password, session)
     await auth.do_auth(store=False)
-    appliance_manager = AppliancesManager(backend_selector, auth)
+    appliance_manager = AppliancesManager(backend_selector, auth=auth, session=session)
```
abmantis, line 59:
```suggestion
    appliance_manager = AppliancesManager(backend_selector, auth, session)
```
we are not missing any arguments, so we do not need to have keyword args here

cli.py
```diff
@@ -51,38 +52,46 @@ def attr_upd():
...
             return
 
     for mo_data in appliance_manager.ovens:
         if mo_data["SAID"] == args.said:
             await show_oven_menu(backend_selector, auth, args.said)
+            await session.close()
             return
 
 
+asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```
abmantis, line 96: Why is this needed? It seems that it is windows only, so we cannot have it here...

cli.py
```diff
@@ -51,38 +52,46 @@ def attr_upd():
...
+    session = aiohttp.ClientSession()
 
-    auth = Auth(backend_selector, args.email, args.password)
+    auth = Auth(backend_selector, args.email, args.password, session)
     await auth.do_auth(store=False)
-    appliance_manager = AppliancesManager(backend_selector, auth)
+    appliance_manager = AppliancesManager(backend_selector, auth=auth, session=session)
     if not await appliance_manager.fetch_appliances():
         logger.error("Could not fetch appliances")
+        await session.close()
```
abmantis, line 62: instead of having to keep track of where we close the session, could we instead use `with` like:
`async with aiohttp.ClientSession() as session:` and just put all the code inside that block?

abmantis (changes requested):
Sorry for the delay in review again.
I left just some minor suggestions and this should be ready to merge!

cli.py
mkmer: I don't believe it breaks anything on linux/mac clients, but it fixes a runtime error in windows.

tests/test_aircon.py
abmantis: That is weird because I got them locally. Maybe it is because you are using windows or something like that.

cli.py
```diff
@@ -52,37 +53,39 @@ def attr_upd():
...
+            print(appliance_manager.washer_dryers)
+            print(appliance_manager.ovens)
             return
 
-    for mo_data in appliance_manager.ovens:
-        if mo_data["SAID"] == args.said:
-            await show_oven_menu(backend_selector, auth, args.said)
+        if not args.said:
+            logger.error("No appliance specified")
+            await session.close()
```
abmantis, line 72:
```suggestion
```

abmantis (changes requested):

tests/test_aircon.py
```diff
@@ -110,6 +113,7 @@ async def test_attributes(
     assert aircon.get_quiet_mode() is True
     await aircon.disconnect()
 
+    appliance_http_client_mock.close_session()
```
abmantis, line 116:
```suggestion
    await appliance_http_client_mock.close_session()
```
I just noticed that this needs to be awaited. Can you please also update the others? 
You can see the warnings for this PR here: https://github.com/abmantis/whirlpool-sixth-sense/actions/runs/3699881981/jobs/6267722655#step:5:132

abmantis (changes requested):

tests/test_aircon.py
mkmer: I can see them at the link - what's strange is running the tests locally in Visual Studio are not throwing these... 
I'll have to dig into my local environment to see why.

It appears that running pytest from the command line does show the errors, but in Visual Studio testing "passes".  I'll have to remember to use the command line (for now)

abmantis: Yeah, the tests pass anyway. Those are just warnings.

abmantis (approved):

abmantis:
Thanks again for your contribution!

PR #32, opened 2023-01-02
Add Unregister_attr_callback

mkmer:
Need to unregister callbacks when entities are removed_from_has()
Need to create a release to fix in HA

### DIFF
--- a/whirlpool/appliance.py
+++ b/whirlpool/appliance.py
@@ -43,7 +43,7 @@ class Appliance:
             self._attr_changed.remove(update_callback)
             LOGGER.debug("Unregistered attr callback")
         except ValueError:
-            LOGGER.debug("Unregistered attr callback not found")
+            LOGGER.error("Attr callback not found")
 
     def _event_socket_handler(self, msg):
         json_msg = json.loads(msg)
--- a/whirlpool/appliance.py
+++ b/whirlpool/appliance.py
@@ -23,6 +23,7 @@ class Appliance:
         backend_selector: BackendSelector,
         auth: Auth,
         said: str,
+        session: aiohttp.ClientSession,
     ):
         self._backend_selector = backend_selector
         self._auth = auth
@@ -31,7 +32,7 @@ class Appliance:
 
         self._data_dict = None
 
-        self._session: aiohttp.ClientSession = None
+        self._session: aiohttp.ClientSession = session
 
     def register_attr_callback(self, update_callback: Callable):
         self._attr_changed.append(update_callback)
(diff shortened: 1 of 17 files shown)


whirlpool/eventsocket.py
```diff
@@ -125,7 +125,6 @@ async def _run(self):
                             continue
                         self._msg_listener("{" + match[0] + "}")
         except (
-            aiohttp.ClientConnectionError,
```
abmantis, line 128: Is this related to the unregister callback?

whirlpool/appliance.py
```diff
@@ -38,6 +38,13 @@ def register_attr_callback(self, update_callback: Callable):
         self._attr_changed.append(update_callback)
         LOGGER.debug("Registered attr callback")
 
+    def unregister_attr_callback(self, update_callback: Callable):
+        try:
+            self._attr_changed.remove(update_callback)
+            LOGGER.debug("Unregistered attr callback")
+        except ValueError:
+            LOGGER.debug("Unregistered attr callback not found")
```
abmantis, line 46:
```suggestion
            LOGGER.error("Attr callback not found")
```

whirlpool/appliance.py
```diff
@@ -151,6 +158,11 @@ async def disconnect(self):
 
     async def start_event_listener(self):
         await self.fetch_data()
+        if self._event_socket != None:
+            await self._event_socket.stop()
+            self._event_socket.start()
+            return
```
abmantis, line 164: Is this related to the unregister callback? also, what is the advantage of reusing the "old" _event_socket?

abmantis (changes requested):

whirlpool/eventsocket.py
mkmer: No - but it was pointed out that the original ClientError is all that is needed as ClientConnectionError is under ClientError.  Thought I might as well clean it up.

whirlpool/appliance.py
mkmer: I suppose not, but it is part of the flow in HA that involves unregister_attr.
I've been chasing a scenario where we were somehow getting 2 event_sockets running - this is the "key" to stopping that behavior.  I suppose we could just stop the existing one then create a new one - or maybe just not stop/start at all if one exists. 
If we kill the existing one and create a new one, it is possible that we lose some registered entities - probably not good.

abmantis: This looks a bit like a workaround to an issue that should be solved on HA. 
Maybe we should instead do `self._event_socket = None` on `stop_event_listener()` and here on `start_event_listener()` we could just do:
```
        if self._event_socket != None:
            LOGGER.error("Event listener already started")
            return
```
Then if this is happening on HA, it should be fixed there to avoid calling start() multiple times.

mkmer: I'm not sure on the HA side how to wait unit the last entity is removed before firing off the disconnect.  Also, how not to connect again when it's already connected...

abmantis: If you only connect() when the platform initializes, and never disconnect(), then this should not be an issue, or is it? There are no "double" connects in that case.

In the future, if you need to use the same Appliance for multiple platforms (say, sensor + switch) you can move the init of Appliances and the connect() call up to the component.

mkmer: So you want me to take it out?
Not reusing the existing socket will result in loss of registered entities.

abmantis: > So you want me to take it out? 

Yeah, I think we should remove and just log an error, otherwise we can be hiding a possible issue on HA side.

> I'm not exactly sure why we end up with 2 eventsockets. I know this bit of code stops it from happening, and preserves the registered entities. I haven't moved the connect() code, it's been in the platform init since day 1.

Maybe they were for two appliances? Anyway, if that happens we should try to understand why and fix it.

mkmer: I think this what you had in mind.

abmantis (approved):
Thanks!

PR #33, opened 2023-01-10
Eventsocket use external session

mkmer:
Noticed we missed using the external session for the eventsocket - probably should release this too. (Not super critical)


abmantis (approved):

abmantis:
0.18.2 is released now

PR #35, opened 2023-02-26
Appliance Unregister test for none, retry on data fetch if auth is required.

mkmer:
Test for None before attempting to use dictionary functions.
Call the callback before registering for event changes (generates new key and headers)
Retry (3 times) on fetch data if auth is required.

### DIFF
--- a/whirlpool/appliance.py
+++ b/whirlpool/appliance.py
@@ -148,8 +148,8 @@ class Appliance:
     def has_attribute(self, attribute):
         if self._data_dict is None:
             LOGGER.error("No data available")
-            return None
-        return attribute in self._data_dict.get("attributes")
+            return False
+        return attribute in self._data_dict.get("attributes", {})
 
     def bool_to_attr_value(self, b: bool):
         return SETVAL_VALUE_ON if b else SETVAL_VALUE_OFF
--- a/whirlpool/appliance.py
+++ b/whirlpool/appliance.py
@@ -15,6 +15,7 @@ ATTR_ONLINE = "Online"
 
 SETVAL_VALUE_OFF = "0"
 SETVAL_VALUE_ON = "1"
+RETRY_COUNT = 3
 
 
 class Appliance:
@@ -41,14 +42,11 @@ class Appliance:
 
     def unregister_attr_callback(self, update_callback: Callable):
         """Unregister callback function."""
-        if self._attr_changed:
-            try:
-                self._attr_changed.remove(update_callback)
-                LOGGER.debug("Unregistered attr callback")
-            except ValueError:
-                LOGGER.error("Attr callback not found")
-        else:
-            LOGGER.error("_attr_changed is None when unregistering callback")
+        try:
+            self._attr_changed.remove(update_callback)
+            LOGGER.debug("Unregistered attr callback")
+        except ValueError:
+            LOGGER.error("Attr callback not found")
 
     def _event_socket_handler(self, msg):
         json_msg = json.loads(msg)
@@ -101,7 +99,7 @@ class Appliance:
             return False
 
         uri = f"{self._backend_selector.base_url}/api/v1/appliance/{self._said}"
-        for n in range(3):
+        for n in range(RETRY_COUNT):
             async with async_timeout.timeout(30):
                 async with self._session.get(uri, headers=self._create_headers()) as r:
                     if r.status == 200:
@@ -125,7 +123,7 @@ class Appliance:
             "body": attributes,
             "header": {"said": self._said, "command": "setAttributes"},
         }
-        for n in range(3):
+        for n in range(RETRY_COUNT):
             async with async_timeout.timeout(30):
                 async with self._session.post(
                     uri, json=cmd_data, headers=self._create_headers()
--- a/whirlpool/appliance.py
+++ b/whirlpool/appliance.py
@@ -52,7 +52,7 @@ class Appliance:
     def _event_socket_handler(self, msg):
         json_msg = json.loads(msg)
         timestamp = json_msg["timestamp"]
-        for (attr, val) in json_msg["attributeMap"].items():
+        for attr, val in json_msg["attributeMap"].items():
             if not self.has_attribute(attr):
                 continue
             self._set_attribute(attr, str(val), timestamp)
--- a/whirlpool/appliance.py
+++ b/whirlpool/appliance.py
@@ -132,7 +132,7 @@ class Appliance:
         return self._data_dict["attributes"][attribute]["value"]
 
     def has_attribute(self, attribute):
-        return attribute in self._data_dict["attributes"]
+        return attribute in self._data_dict.get("attributes")
 
     def bool_to_attr_value(self, b: bool):
         return SETVAL_VALUE_ON if b else SETVAL_VALUE_OFF
--- a/whirlpool/appliance.py
+++ b/whirlpool/appliance.py
@@ -144,7 +144,7 @@ class Appliance:
         return self._data_dict["attributes"][attribute]["value"]
 
     def has_attribute(self, attribute):
-        if self._data_dict:
+        if self._data_dict is not None:
             return attribute in self._data_dict.get("attributes")
         return None
 
--- a/whirlpool/appliance.py
+++ b/whirlpool/appliance.py
@@ -106,9 +106,10 @@ class Appliance:
                         self._data_dict = json.loads(await r.text())
                         return True
                     elif r.status == 401:
+                        LOGGER.error("Fetching data failed (%s). Doing reauth", r.status)
                         await self._auth.do_auth()
-
-        LOGGER.error("Fetching data failed (%s)", r.status)
+                    else:
+                        LOGGER.error("Fetching data failed (%s)", r.status)
         return False
 
     async def send_attributes(self, attributes):
--- a/whirlpool/appliance.py
+++ b/whirlpool/appliance.py
@@ -1,8 +1,8 @@
-import aiohttp
-import async_timeout
 import logging
 import json
 from typing import Callable
+import async_timeout
+import aiohttp
 
 from .backendselector import BackendSelector
 
@@ -19,6 +19,8 @@ REQUEST_RETRY_COUNT = 3
 
 
 class Appliance:
+    """Whirlpool appliance class"""
+
     def __init__(
         self,
         backend_selector: BackendSelector,
@@ -91,9 +93,11 @@ class Appliance:
 
     @property
     def said(self):
+        """Return Appliance SAID"""
         return self._said
 
     async def fetch_data(self):
+        """Fetch appliance data from web api"""
         if not self._session:
             LOGGER.error("Session not started")
             return False
@@ -114,26 +118,8 @@ class Appliance:
                         LOGGER.error("Fetching data failed (%s)", r.status)
         return False
 
-    async def check_login(self):
-        """Verify user is logged in, reauth if not logged in"""
-        if not self._session:
-            LOGGER.error("Session not started")
-            return False
-
-        uri = f"{self._backend_selector.base_url}/api/v1/appliance/{self._said}"
-        for retry in range(REQUEST_RETRY_COUNT):
-            async with async_timeout.timeout(30):
-                async with self._session.get(uri, headers=self._create_headers()) as r:
-                    if r.status == 200:
-                        return True
-                    elif r.status == 401:
-                        LOGGER.error("Login check failed (%s). Doing reauth", r.status)
-                        await self._auth.do_auth()
-                    else:
-                        LOGGER.error("Login failed (%s)", r.status)
-        return False
-
     async def send_attributes(self, attributes):
+        """Send attributes to appliance api"""
         if not self._session:
             LOGGER.error("Session not started")
             return False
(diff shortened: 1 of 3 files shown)


whirlpool/appliance.py
```diff
@@ -35,20 +35,25 @@ def __init__(
...
         LOGGER.debug("Registered attr callback")
 
     def unregister_attr_callback(self, update_callback: Callable):
-        try:
-            self._attr_changed.remove(update_callback)
-            LOGGER.debug("Unregistered attr callback")
-        except ValueError:
-            LOGGER.error("Attr callback not found")
+        """Unregister callback function."""
+        if self._attr_changed:
```
abmantis, line 44: in what situations can this be None? It is inited in __init__ and is never set to None, so I believe we should not need this check

whirlpool/appliance.py
```diff
@@ -96,15 +101,16 @@ async def fetch_data(self):
...
-        async with async_timeout.timeout(30):
-            async with self._session.get(uri, headers=self._create_headers()) as r:
-                self._data_dict = json.loads(await r.text())
-                if r.status == 200:
-                    return True
-                elif r.status == 401:
-                    await self._auth.do_auth()
-
-                LOGGER.error(f"Fetching data failed ({r.status})")
+        for n in range(3):
```
abmantis, line 104:
```suggestion
        try_count = 3
        for n in range(retry_count):
```
just to avoid magic numbers

whirlpool/appliance.py
```diff
@@ -96,15 +101,16 @@ async def fetch_data(self):
...
+            async with async_timeout.timeout(30):
+                async with self._session.get(uri, headers=self._create_headers()) as r:
+                    if r.status == 200:
+                        self._data_dict = json.loads(await r.text())
+                        return True
+                    elif r.status == 401:
+                        await self._auth.do_auth()
+
+        LOGGER.error("Fetching data failed (%s)", r.status)
```
abmantis, line 113:
```suggestion
                    elif r.status == 401:
                        LOGGER.error("Fetching data failed (%s). Doing reauth", r.status)
                        await self._auth.do_auth()
                    else:
                        LOGGER.error("Fetching data failed (%s)", r.status)
```

whirlpool/appliance.py
```diff
@@ -139,7 +145,9 @@ def get_attribute(self, attribute):
         return self._data_dict["attributes"][attribute]["value"]
 
     def has_attribute(self, attribute):
-        return attribute in self._data_dict.get("attributes")
+        if self._data_dict is not None:
+            return attribute in self._data_dict.get("attributes")
+        return None
```
abmantis, line 150:
```suggestion
        if self._data_dict is None:
            LOGGER.error("No data available")
            return None
        return attribute in self._data_dict.get("attributes")
```

abmantis (changes requested):

whirlpool/appliance.py
mkmer: Should we make it a parameter (aka TRY_COUNT = 3 in the top of the file?) Probably should do the same to the write function... since I just copied it form there.

whirlpool/appliance.py
mkmer: You may be correct here - I put this test in the "wrong place" trying to stop the NoneType iteration error.

mkmer:
Still at a bit of a loss to understand why this started acting up.  It was running 100% for 3 months while we waited for reviews.  
Maybe a quick review on what we are doing to keep the eventsocket running is in order?
Additionally, I copied this event socket code in AIOAladdinConnect which has been solid for 6+ months.  Obviously I've missed something....

whirlpool/eventsocket.py
```diff
@@ -62,95 +62,96 @@ async def _recv_msg(self, websocket: aiohttp.ClientWebSocketResponse):
...
+                LOGGER.debug(f"Connecting to {self._url}")
+                async with self._session.ws_connect(
+                    self._url,
+                    timeout=timeout,
+                    autoclose=True,
+                    autoping=True,
+                    heartbeat=45,
+                ) as ws:
+                    self._websocket = ws
+                    await self._con_up_listener()
```
abmantis, line 78: why move this up?

whirlpool/eventsocket.py
```diff
@@ -62,95 +62,96 @@ async def _recv_msg(self, websocket: aiohttp.ClientWebSocketResponse):
...
-
-            LOGGER.info(
-                f"Waiting to reconnect short delay {RECONNECT_SHORT_DELAY} seconds"
-            )
-            await asyncio.sleep(RECONNECT_SHORT_DELAY)
-
-            LOGGER.info("Reconnecting...")
-            self._run_future = asyncio.get_event_loop().create_task(self._run())
+                LOGGER.info("Reconnecting...")
+                # self._run_future = asyncio.get_event_loop().create_task(self._run())
```
abmantis, line 154:
```suggestion
```

whirlpool/auth.py
```diff
@@ -101,15 +102,15 @@ async def _do_auth(self, refresh_token):
                     return await self._do_auth(refresh_token=None)
                 return None
 
-    async def do_auth(self, store=True):
+    async def do_auth(self, store: bool = False) -> bool | None:
```
abmantis, line 105:
```suggestion
    async def do_auth(self, store: bool = False) -> bool:
```

whirlpool/auth.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import annotations
```
abmantis, line 1: why is this needed for?

whirlpool/appliance.py
```diff
@@ -15,6 +15,7 @@
 
 SETVAL_VALUE_OFF = "0"
 SETVAL_VALUE_ON = "1"
+RETRY_COUNT = 3
```
abmantis, line 18:
```suggestion
REQUEST_RETRY_COUNT = 3
```
Just to make it a bit more explicit

abmantis (changes requested):

whirlpool/eventsocket.py
mkmer: This is the loop change - moved to a while loop, removed the thread restart logic - aka just keep looping until `not self._running`  There's really no need to let the thread die then restart it.
The `stop()` logic sets self._running = False to cause this thread to end.

mkmer: Sorry, I moved that to trigger a login if it gets a 401.  With it in the original location, we can get an error and never try a login.

whirlpool/auth.py
mkmer: Typing

whirlpool/eventsocket.py
abmantis: Sorry, I am not sure I understand. I think `_con_up_listener()` should only be called after the connect and subscribe messages are sent, since that is effectively when the connection is ready.
This will loop if we get an error, right?

whirlpool/auth.py
abmantis: I removed it and got no errors. What typing do you refer to that requires this?

whirlpool/eventsocket.py
mkmer: `_con_up_listener()' is `fetch_data()` - https://github.com/abmantis/whirlpool-sixth-sense/blob/26d8cecbec6f0ff2e6a83efa8db54b3c8197a3d3/whirlpool/appliance.py#L93-L108
Calling fetch data first, makes sure we are logged in before attempting to register the callacks via the code that follows.  I witnessed this occurring so I moved this up.  I'm not sure how we actually became logged out before attempting to register, but it happened.

whirlpool/auth.py
mkmer: Any time typing is used say  `-> int | None` you need the annotations import or it will complain.
Maybe we removed the one I had that needed it.

whirlpool/auth.py
```diff
@@ -57,7 +58,7 @@ def _save_auth_data(self):
         with open(AUTH_JSON_FILE, "w") as f:
             json.dump(self._auth_dict, f)
 
-    async def _do_auth(self, refresh_token):
+    async def _do_auth(self, refresh_token: str) -> str | None:
```
mkmer, line 61: This is an example - if you don't have `from __future__ import annotations` this will not work.

whirlpool/auth.py
mkmer: https://dev.to/tiangolo/the-future-of-fastapi-and-pydantic-is-bright-3pbm

whirlpool/appliance.py
mkmer: 2076 Traceback (most recent call last):
2077   File "/usr/src/homeassistant/homeassistant/config_entries.py", line 539, in async_unload
2078     result = await component.async_unload_entry(hass, self)
2079   File "/usr/src/homeassistant/homeassistant/components/sensor/__init__.py", line 137, in async_unload_entry
2080     return await component.async_unload_entry(entry)
2081   File "/usr/src/homeassistant/homeassistant/helpers/entity_component.py", line 186, in async_unload_entry
2082     await platform.async_reset()
2083   File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 745, in async_reset
2084     await asyncio.gather(*tasks)
2085   File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 845, in async_remove
2086     await self.async_will_remove_from_hass()
2087   File "/usr/src/homeassistant/homeassistant/components/whirlpool/sensor.py", line 270, in async_will_remove_from_hass
2088     await self._wd.disconnect()
2089   File "/usr/local/lib/python3.10/site-packages/whirlpool/appliance.py", line 157, in disconnect
2090     await self.stop_event_listener()
2091   File "/usr/local/lib/python3.10/site-packages/whirlpool/appliance.py", line 175, in stop_event_listener
2092     await self._event_socket.stop()
2093   File "/usr/local/lib/python3.10/site-packages/whirlpool/eventsocket.py", line 166, in stop
2094     await self._run_future
2095   File "/usr/local/lib/python3.10/site-packages/whirlpool/eventsocket.py", line 125, in _run
2096     self._msg_listener("{" + match[0] + "}")
2097   File "/usr/local/lib/python3.10/site-packages/whirlpool/appliance.py", line 52, in _event_socket_handler
2098     if not self.has_attribute(attr):
2099   File "/usr/local/lib/python3.10/site-packages/whirlpool/appliance.py", line 142, in has_attribute
2100     return attribute in self._data_dict.get("attributes")
2101 TypeError: argument of type 'NoneType' is not iterable

whirlpool/eventsocket.py
abmantis: > `_con_up_listener()` is `fetch_data()` -

Yes, I know. But the PoV of this class (Eventsocket), `_con_up_listener` can be anything. On Eventsocket we should not assume that the callback does "X".
> 
> https://github.com/abmantis/whirlpool-sixth-sense/blob/26d8cecbec6f0ff2e6a83efa8db54b3c8197a3d3/whirlpool/appliance.py#L93-L108
> 
> Calling fetch data first, makes sure we are logged in before attempting to register the callbacks via the code that follows. I witnessed this occurring so I moved this up. I'm not sure how we actually became logged out before attempting to register, but it happened.
> 
> Calling `fetch_data()` after the change of state is registered was intended to make sure the state machine is updated before switch to 100% change of state. By calling it "just before" registering there is a small chance we miss a state change, but it's better than not registering due to a login error and being 100% off until the integration is reloaded.

But Eventsocket does not know that, so the callback should be generic anyway. Otherwise, if in the future `fetch_data()` changes and does not login for some reason, then this assumption will be broken. Also, someone working on this later will also not know that we had this assumption here, by looking at Eventsocket.
So, if we need to ensure that auth is done before, that should be handled entirely here without relying on the callback to do it, IMO.

whirlpool/auth.py
abmantis: I don't think it is needed anymore, unless you are using an old python version. Since HA uses 3.10 as minimum, we should also do the same here. Maybe your python env was still using a previous version?

whirlpool/auth.py
abmantis: I have no issues here on that line

whirlpool/appliance.py
abmantis: Ah, so the attributes key is missing.
I'll add a line suggestion to this PR to fix that.

whirlpool/appliance.py
```diff
@@ -139,6 +146,9 @@ def get_attribute(self, attribute):
         return self._data_dict["attributes"][attribute]["value"]
 
     def has_attribute(self, attribute):
+        if self._data_dict is None:
+            LOGGER.error("No data available")
+            return None
         return attribute in self._data_dict.get("attributes")
```
abmantis, line 152:
```suggestion
        return attribute in self._data_dict.get("attributes", {})
```

whirlpool/appliance.py
```diff
@@ -139,6 +146,9 @@ def get_attribute(self, attribute):
         return self._data_dict["attributes"][attribute]["value"]
 
     def has_attribute(self, attribute):
+        if self._data_dict is None:
+            LOGGER.error("No data available")
+            return None
```
abmantis, line 151:
```suggestion
            return False
```
lets keep this func boolean

abmantis (changes requested):

whirlpool/eventsocket.py
mkmer: So what's the answer? Add a login function here if we get a 401?  It can't just "die" because were not logged at this point.

whirlpool/auth.py
mkmer: According to the link, it didn't make it into 3.10.

mkmer:
Python 3.9 test now fails without `from __future__ import annotations` - are we going to stop running this test?

whirlpool/auth.py
mkmer: This fails the Python 3.9 tests without it...

whirlpool/eventsocket.py
mkmer: I've added a `check_login` function & callback - is this what you had in mind?

abmantis:
> Python 3.9 test now fails without `from __future__ import annotations` - are we going to stop running this test?

I've updated the jobs. Please update from master and this should now be fixed.

abmantis (changes requested):

whirlpool/eventsocket.py
```diff
@@ -62,94 +64,96 @@ async def _recv_msg(self, websocket: aiohttp.ClientWebSocketResponse):
...
+                LOGGER.debug(f"Connecting to {self._url}")
+                async with self._session.ws_connect(
+                    self._url,
+                    timeout=timeout,
+                    autoclose=True,
+                    autoping=True,
+                    heartbeat=45,
+                ) as ws:
+                    self._websocket = ws
+                    await self._check_login_listener()
```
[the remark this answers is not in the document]
mkmer: This looks "ok" - I will need to run it for a bit to make sure it's handling the real world as we expect.  The original relocation has been running locally for a few months and covered the issues.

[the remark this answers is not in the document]
mkmer: Been running and "recovering" for 3 days - looks like we're good to go with this change.

abmantis (approved):

PR #36, opened 2023-02-27
Lint Cleanup and lazy format for logging

mkmer:
Looks like we need to do lazy format for logging.
While fixing the logging, did some lint requested updates (doc strings, `is not None` vs `!= None` etc)

### DIFF
--- a/whirlpool/appliance.py
+++ b/whirlpool/appliance.py
@@ -52,7 +52,7 @@ class Appliance:
     def _event_socket_handler(self, msg):
         json_msg = json.loads(msg)
         timestamp = json_msg["timestamp"]
-        for (attr, val) in json_msg["attributeMap"].items():
+        for attr, val in json_msg["attributeMap"].items():
             if not self.has_attribute(attr):
                 continue
             self._set_attribute(attr, str(val), timestamp)
--- a/whirlpool/appliance.py
+++ b/whirlpool/appliance.py
@@ -132,7 +132,7 @@ class Appliance:
         return self._data_dict["attributes"][attribute]["value"]
 
     def has_attribute(self, attribute):
-        return attribute in self._data_dict["attributes"]
+        return attribute in self._data_dict.get("attributes")
 
     def bool_to_attr_value(self, b: bool):
         return SETVAL_VALUE_ON if b else SETVAL_VALUE_OFF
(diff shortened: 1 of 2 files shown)


whirlpool/appliance.py
```diff
@@ -35,20 +37,24 @@ def __init__(
...
         LOGGER.debug("Registered attr callback")
 
     def unregister_attr_callback(self, update_callback: Callable):
-        try:
-            self._attr_changed.remove(update_callback)
-            LOGGER.debug("Unregistered attr callback")
-        except ValueError:
-            LOGGER.error("Attr callback not found")
+        """Unregister callback function."""
+        if self._attr_changed:
```
abmantis, line 46: there is no way this can be None so I don't think this check is needed

abmantis (changes requested):

whirlpool/appliance.py
```diff
@@ -60,14 +66,13 @@ def _create_headers(self):
...
             "Authorization": "Bearer " + self._auth.get_access_token(),
             "Content-Type": "application/json",
-            # "Host": "api.whrcloud.eu",
             "User-Agent": "okhttp/3.12.0",
             "Pragma": "no-cache",
             "Cache-Control": "no-cache",
         }
 
     def _set_attribute(self, attribute, value, timestamp):
-        LOGGER.debug(f"Updating attribute {attribute} with {value} ({timestamp})")
```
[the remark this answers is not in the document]
mkmer: Lint says so.... I believe the logs also show {} instead of the value.
If we don't need to, we need to fix the lint rules.

whirlpool/appliance.py
mkmer: It can be, if you unregistered multiple times it may be None, if we don't test for it, it raises an error.

abmantis: It is never set to None. If you mean unregister being called when the array is empty, then the "Attr callback not found" error should indeed be logged, because that should not happen and the client (HA?) needs to fix that.

whirlpool/appliance.py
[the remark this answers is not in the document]
abmantis: I just ran `black` and this gave me no errors.

abmantis (changes requested):
I just ran `black` and it only changed some minor stuff. I already opened a PR and merged it to master. The check job is passing. Do you want to clean this one to keep only the docstring changes?

PR #39, opened 2023-04-08
Add model and serial numbers

austinmroczek:
Retain the appliance model number and serial number when loading location info.  I used dict.get() in case other appliances don't provide these values.

This could later be added to Home Assistant for a more complete `DeviceInfo`.

I didn't add specific tests, but existing tests pass.


abmantis:
Thanks! Please update the branch with master, since it fixes the lint

austinmroczek:
updated, thanks

PR #44, opened 2023-06-26
Eventsocket read timeout and poll update

mkmer:
In order to better handle recent failures in the Whirlpool API to send state changes related to https://github.com/home-assistant/core/issues/95177, this PR adds a read timeout and code to restart the socket, refresh the data(via `self._con_up_listener()`) after ~60 minutes of no activity on the event socket.
Also, while adding this feature, noticed that ` fetch_data(self)` was not triggering a callback to HA notifying HA of an update. Without this notification, restarting the `eventsocket()` did not update HA as expected.

If/When Whirlpool fixes the problems, this code will not trigger as a "functioning" event socket updates the "operating time" every 60 minutes.


whirlpool/appliance.py
```diff
@@ -51,12 +51,13 @@ def unregister_attr_callback(self, update_callback: Callable):
...
-        for attr, val in json_msg["attributeMap"].items():
-            if not self.has_attribute(attr):
-                continue
-            self._set_attribute(attr, str(val), timestamp)
+        if msg is not None:
+            json_msg = json.loads(msg)
+            timestamp = json_msg["timestamp"]
+            for attr, val in json_msg["attributeMap"].items():
+                if not self.has_attribute(attr):
+                    continue
```
mkmer, line 59: This was a mistake from the beginning - we need to tell HA to update (via calback) after we sync the data.  The base code is only updating the internal data.

whirlpool/eventsocket.py
```diff
@@ -133,9 +148,10 @@ async def _run(self):
                             connected_msg_done = True
                             continue
 
-                        if not subscribe_msg_done:
+                        elif not subscribe_msg_done:
                             subscribe_msg_done = True
                             await self._con_up_listener()
+                            self._msg_listener(None)
```
mkmer, line 154: This is part of the "original" mistake - needing to tell HA new data is available.

whirlpool/appliance.py
mkmer: I since moved this around a bit - rather than calling the update with (None), lets just update HA whenever a fetch occurs.  That way it just "works" whenever a fetch is triggered.

whirlpool/eventsocket.py
mkmer: See previous comment....

whirlpool/appliance.py
```diff
@@ -71,6 +75,13 @@ def _create_headers(self):
             "Cache-Control": "no-cache",
         }
 
+    def _check_previous_attribute_value(self, attr, json_msg):
+        """Confirm current stored value is equal to expected previous value."""
+        return (
+            str(json_msg["processedValueMap"]["previousValueMap"][attr])
+            == self._data_dict["attributes"][attr]["value"]
+        )
+
```
mkmer, line 84: Besides the eventsocket stopping, it also is exhibiting failure to send "some" messages then starting back up on it's own.  Found that the messages contains a previous value we can use to see if we are up to date - if not, just refresh everything with a fetch.

whirlpool/appliance.py
```diff
@@ -56,7 +56,11 @@ def _event_socket_handler(self, msg):
         for attr, val in json_msg["attributeMap"].items():
             if not self.has_attribute(attr):
                 continue
-            self._set_attribute(attr, str(val), timestamp)
+            if self._check_previous_attribute_value(attr, json_msg):
+                self._set_attribute(attr, str(val), timestamp)
+            else:
+                LOGGER.debug("Previous data mismatch - fetching new dataset")
+                self.fetch_data()
```
abmantis, line 63: we cannot do this since it will do one request for each attribute.
I also don't really think it is a good approach to be making requests just because we got unchanged data. If the data is not changing, then we will keep requesting data non stop.

abmantis (changes requested):
I would prefer to wait a bit to see if whirlpool fixes the issue (since it also affects the app), before trying to work around it.

mkmer:
Ok. I will separate the mistake from this pr, HA needs to be notified after the socket reconnects to update its data. Currently it is not, and it takes a state change message to trigger the update.

whirlpool/appliance.py
mkmer: It will only do one request for each attribute that is mismatched - after we do the `fetch_data()` request 1 time, the previous data will be matched for every attribute, assuming each attribute had a mismatch.

mkmer:
I have been running this code - unfortunately, it doesn't help much :(
If we miss an attribute update,  a different attribute comes along with an update that is not mis-matched.  The attribute that is mismatched is "long finished" and no longer sends updates for quite a while (in many cases. the next load of laundry).

Maybe we just need to do a periodic update on HA side just to be sure things are updated for now....

PR #45, opened 2023-07-10
Notify HA of updates on fetch_data()

mkmer:
Need to tell listeners when new data is fetched - found when restarting the socket and calling fetch_data() HA was not notified of updated entities until an event message was received.

### DIFF
--- a/whirlpool/eventsocket.py
+++ b/whirlpool/eventsocket.py
@@ -132,7 +132,7 @@ class EventSocket:
                         if not connected_msg_done:
                             connected_msg_done = True
                             continue
-                        
+
                         elif not subscribe_msg_done:
                             subscribe_msg_done = True
                             await self._con_up_listener()
--- a/whirlpool/eventsocket.py
+++ b/whirlpool/eventsocket.py
@@ -133,7 +133,7 @@ class EventSocket:
                             connected_msg_done = True
                             continue
 
-                        elif not subscribe_msg_done:
+                        if not subscribe_msg_done:
                             subscribe_msg_done = True
                             await self._con_up_listener()
                             continue
--- a/whirlpool/eventsocket.py
+++ b/whirlpool/eventsocket.py
@@ -132,8 +132,8 @@ class EventSocket:
                         if not connected_msg_done:
                             connected_msg_done = True
                             continue
-
-                        if not subscribe_msg_done:
+                        
+                        elif not subscribe_msg_done:
                             subscribe_msg_done = True
                             await self._con_up_listener()
                             continue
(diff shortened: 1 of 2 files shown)


abmantis:
I am not sure I understand the need for this change. `fetch_data()` is only called on startup when HA calls `connect()`, so there is no need to call the attr changed callback there.

mkmer:
It is also called when the socket reconnects:
https://github.com/abmantis/whirlpool-sixth-sense/blob/f217ca550a275e1a04fd567a8efd092ad2968bce/whirlpool/appliance.py#L192C8-L192C8
which is mapped to this:
https://github.com/abmantis/whirlpool-sixth-sense/blob/f217ca550a275e1a04fd567a8efd092ad2968bce/whirlpool/eventsocket.py#L138

When the event socket reconnects, we don't "know" the states for sure, so we do a fetch to make sure everything is up to date.  Unfortunately, we don't tell HA that there is new data to update when this happens.

This will also allow us to drive 1 timed update function and still trigger all other entities (The bandaid for a crappy event socket)

whirlpool/eventsocket.py
```diff
@@ -133,7 +133,7 @@ async def _run(self):
                             connected_msg_done = True
                             continue
 
-                        if not subscribe_msg_done:
```
mkmer, line 136: I forgot about this one - found it's not waiting for the connected_msg_done before sending the subscribe_msg_done.

abmantis: this elif is not needed, since the if block has `continue` already

abmantis:
> It is also called when the socket reconnects: https://github.com/abmantis/whirlpool-sixth-sense/blob/f217ca550a275e1a04fd567a8efd092ad2968bce/whirlpool/appliance.py#L192C8-L192C8 which is mapped to this:
> 
> https://github.com/abmantis/whirlpool-sixth-sense/blob/f217ca550a275e1a04fd567a8efd092ad2968bce/whirlpool/eventsocket.py#L138
> 
> When the event socket reconnects, we don't "know" the states for sure, so we do a fetch to make sure everything is up to date. Unfortunately, we don't tell HA that there is new data to update when this happens.
> 
> This will also allow us to drive 1 timed update function and still trigger all other entities (The bandaid for a crappy event socket)

Ah I see! Makes sense!

whirlpool/eventsocket.py
mkmer: True - I thought I had traced this through and it was behaving differently, but on second review it was working as expected.

abmantis (approved):

abmantis:
Thanks!

mkmer:
Can we get a release in pypi for this? Looking to activate the polling feature in HA and this is necessary for HA to update all entities.

abmantis:
@mkmer sure! I just released it as v0.18.4.

PR #49, opened 2024-01-07
Add additional whirlpool_android client credentials, add support for shared devices

NodeJSmith:
This replaces the whirlpool brand client_id and client_secret with updated values that work, as the old ones were no longer valid.

This also adds the methods `_get_owned_appliances`, `_get_shared_appliances`, and `_add_appliance` to `AppliancesManager`. The `fetch_appliances` method will call `_get_owned_appliances` first and then `_get_shared_appliances`, ensuring that both types are added to the appliance list.

There are some other minor changes to fix failing tests, including the addition of a noop async `wait_for_close` method to the `aiohttp` mock and pinning `aiohttp` to below 3.9.0. The `wait_for_close` is actually not required any longer with `aiohttp` pinned to a lower version, but I figure it's better to leave it there to save the next person the headache.


abmantis:
Hi, thanks for your work! 
The current client id/secret work for me, but not the ones you used. Can you share the link for the Android app you are using?

I'll try to review the code next week, but can't promise since it will be a busy one.

NodeJSmith:
@abmantis That's so weird, I only started digging into this because I couldn't authenticate in HA with a username/password I knew was correct. I spent three days trying to figure out the issue before finally being able to authenticate with the new values. I wonder what makes the difference.

The version if 5.9.6, from this link https://www.apkmonk.com/download-app/com.whirlpool.android.wpapp/4_com.whirlpool.android.wpapp_2023-12-13.apk/

NodeJSmith:
@abmantis AFAICT the app has multiple credentials for a given Region + Brand + Environment combo. Both sets of credentials are in there for US + Whirlpool + Production, for example. I haven't checked yet if it's the same for other brands. I also can't tell if there is any way to determine which ones should be used for a given login, nothing I've seen so far indicates that.

NodeJSmith:
@abmantis I pushed a new commit, making the credentials a list of dictionaries and updating the auth process to try them out in sequence, returning as soon as one succeeds. If none succeed it will return None as it does currently, indicating that authentication was not successful. I think this should satisfy both situations - I tested with my login and it returned the 401 response for the initial credentials and then 200 for the second, letting me use the CLI to view my appliances. It should work for you, returning 200 on the first try. Let me know if you have any concerns, but I think this is the best solution given the weird situation.

tests/test_auth.py
```diff
@@ -13,8 +13,8 @@
 
 AUTH_URL = f"{BACKEND_SELECTOR_MOCK.base_url}/oauth/token"
 AUTH_DATA = {
-    "client_id": BACKEND_SELECTOR_MOCK.client_id,
-    "client_secret": BACKEND_SELECTOR_MOCK.client_secret,
+    "client_id": BACKEND_SELECTOR_MOCK.credentials[0]["client_id"],
```
abmantis, line 16: Can you also add tests for the new features, please?
- shared appliances
- ensuring that the auth goes trough the two client credentials?

abmantis (changes requested):
Thanks a lot for your work, and sorry for the delay getting back to this!
I've left just some minor comments, and it would also be really nice if you can add tests for the new stuff!

abmantis:
By the way, clever idea to pool trough the multipe creds! That way we don't need to ask the user anything.

NodeJSmith:
@abmantis thanks for review, I'll get the changes and tests done in the next day or two 👍

NodeJSmith:
@abmantis I got the tests added and made the changes you suggested.

There were a couple of things that got changed slightly in order to better facilitate the tests. I moved the generation of the auth url to `BackendSelector` so that it could be used by both the package and the tests without having to copy the logic into the tests. I also created Enums in the mock `BackendSelector` for brand and region, since those are required for the headers in `AppliancesManager` now.

I also split out some more methods, separating out the creation of the authorization body to a separate method in `Auth` and doing the same with getting the account_id in `AppliancesManager`. Both of those just made it easier to test the independent functionality in different tests.

I added some test data into the tests folder to use in the tests for `AppliancesManager` - I used my own data and masked any sensitive values, but I only have a washer/dryer so it's not very robust as far as what it can test. I'd love to get air conditioner and/or over data in there for better coverage.

Lastly, I added the async-timeout to the requirements so that it's still included in the install for Python >= 3.11 and I changed the type hints for dict/list to import from typing, to support older versions of python

whirlpool/appliancesmanager.py
```diff
@@ -18,23 +18,24 @@ def __init__(
...
         self._backend_selector = backend_selector
         self._auth = auth
-        self._aircons: list[dict[str, Any]] = []
-        self._washer_dryers: list[dict[str, Any]] = []
-        self._ovens: list[dict[str, Any]] = []
+        self._aircons: List[Dict[str, Any]] = []
+        self._washer_dryers: List[Dict[str, Any]] = []
+        self._ovens: List[Dict[str, Any]] = []
         self._session: aiohttp.ClientSession = session
+        self._account_id: str = self._auth._auth_dict.get("accountId")
```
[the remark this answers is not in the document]
abmantis: Also, we should not access _auth_dict directly. Maybe add a get_account_id() to Auth?

whirlpool/auth.py
```diff
@@ -58,49 +59,42 @@ def _save_auth_data(self):
...
-        auth_data = {
-            "client_id": self._backend_selector.client_id,
-            "client_secret": self._backend_selector.client_secret,
-        }
-
+    def _get_auth_body(
+        self, refresh_token: str, creds: Dict[str, str]
+    ) -> Dict[str, str]:
         if refresh_token:
             LOGGER.info("Fetching auth with refresh token")
```
abmantis, line 66:
```suggestion
            LOGGER.info("Using refresh token in auth body")
```
Since this method is not the one doing the fetch anymore, lets update the log too :)

whirlpool/auth.py
```diff
@@ -58,49 +59,42 @@ def _save_auth_data(self):
...
-            auth_data.update(
-                {
-                    "grant_type": "refresh_token",
-                    "refresh_token": refresh_token,
-                }
-            )
+            auth_data = {"grant_type": "refresh_token", "refresh_token": refresh_token}
+
         else:
             LOGGER.info("Fetching auth with user/pass")
```
abmantis, line 70:
```suggestion
            LOGGER.info("Using user/pass in auth body")
```

whirlpool/auth.py
```diff
@@ -58,49 +59,42 @@ def _save_auth_data(self):
...
-            # "WP-CLIENT-COUNTRY": "EN",
-        }
-
-        auth_data = {
-            "client_id": self._backend_selector.client_id,
-            "client_secret": self._backend_selector.client_secret,
-        }
-
+    def _get_auth_body(
+        self, refresh_token: str, creds: Dict[str, str]
```
abmantis, line 63:
```suggestion
        self, refresh_token: str, client_creds: Dict[str, str]
```

pyproject.toml
```diff
@@ -15,8 +15,9 @@ classifiers = [
 requires-python = ">=3.6"
 dependencies = [
     "aioconsole>=0.3.1",
-    "aiohttp>=3.7.2",
+    "aiohttp>=3.9.1",
     "websockets>=8.1",
+    "async-timeout>=4.0.3",
```
abmantis, line 20:
```suggestion
```
I think this is a dependency of aiohttp already, so we do not need to specify it. Did that change?

requirements.txt
```diff
@@ -1,3 +1,4 @@
 aioconsole>=0.3.1
-aiohttp>=3.7.2
+aiohttp>=3.9.1
 websockets>=8.1
+async-timeout>=4.0.3
```
abmantis, line 4:
```suggestion
```

whirlpool/appliancesmanager.py
```diff
@@ -18,71 +18,108 @@ def __init__(
...
+            self._account_id = data["accountId"]
+
+        return self._account_id
+
+    async def fetch_appliances(self):
+        account_id = await self._get_account_id()
+        success_owned = await self._get_owned_appliances(account_id)
+        success_shared = await self._get_shared_appliances()
+
+        return success_owned and success_shared
```
abmantis, line 122:
```suggestion
        return success_owned or success_shared
```
In my case, the shared_appliances API fails with a 404, so we cannot enforce it.

abmantis (changes requested):
I was unable to review the tests yet :( 
I'll try to get back at this tomorrow, hopefully!

pyproject.toml
NodeJSmith: It is no longer a dependency of aiohttp for 3.11 and above, so we have to add it to our own dependencies to ensure we have everything we need for running on those versions.

whirlpool/appliancesmanager.py
NodeJSmith: That's good to know, didn't realize it would return a 404 if there were no shared appliances. I'll update this then

whirlpool/auth.py
NodeJSmith: Renamed, good catch

whirlpool/appliancesmanager.py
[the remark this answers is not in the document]
NodeJSmith: So I added the `get_account_id()` and it's no longer being stored, but as I was doing this I started wondering what the context of this is - if Auth has an access token wouldn't it also have the accountId? Do these not match sometimes?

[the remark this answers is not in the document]
abmantis: I think that when I started implementing this, the accountId was not part of the auth response, so I had to use the other endpoint to get it. Whirlpool broke the API a while ago, and maybe it was introduced at that point.
I just checked and indeed, it is part of the auth now! But I am afraid to remove it completely for now, so I think your implementation is the best of both worlds.

pyproject.toml
abmantis: oh, right! thanks!

whirlpool/appliancesmanager.py
abmantis: I assume that this is because I am using the "old" client credentials or something, so it is not authorized to use the shared appliances endpoint. But who knows :shrug:

tests/test_appliancesmanager.py
```diff
@@ -0,0 +1,148 @@
...
+    get_mock_coro,
+    mock_appliancesmanager_get_account_id_get,
+    mock_appliancesmanager_get_owned_appliances_get,
+    mock_appliancesmanager_get_shared_appliances_get,
+)
+
+BACKEND_SELECTOR_MOCK = BackendSelectorMock()
+
+
+async def test_create_headers_includes_wp_client_brand(
```
abmantis, line 20: I would prefer to avoid testing internal logic/methods of classes. Instead, I think we can have the fetch_appliances() test check that the expected headers are on the http_client_mock.mock_calls arguments.

That way we rely only on public methods + mocks to ensure behavior, making the tests less prone to break (since the public interface of the class should not change as often).

You can check the method `assert_appliance_setter_call` in tests/utils.py on how to assert using the http_client_mock

tests/test_appliancesmanager.py
```diff
@@ -0,0 +1,148 @@
...
+    am = AppliancesManager(BACKEND_SELECTOR_MOCK, auth, http_client_mock.session)
+
+    headers = am._create_headers()
+
+    assert headers["WP-CLIENT-BRAND"] == BACKEND_SELECTOR_MOCK.brand.name
+
+    await http_client_mock.close_session()
+
+
+async def test_account_id_set_if_populated_in_auth_dict(
```
abmantis, line 36: As in the above test, I think we can test this using only the public fetch_appliances().
We could have the "normal" `test_fetch_appliances()` that uses the one from the auth_dict, since that is the expected use case, and another `test_fetch_appliances_no_account_id_on_auth()` or something, that tests the other case

tests/test_appliancesmanager.py
```diff
@@ -0,0 +1,148 @@
...
+
+    account_id = await am._get_account_id()
+
+    assert account_id == "12345"
+    assert account_id != auth.get_account_id()
+
+    await http_client_mock.close_session()
+
+
+async def test_fetch_appliances_calls_both_get_methods(
```
abmantis, line 73: also in this test we can make the check using only the http_client_mock, without accessing internal methods.

tests/test_appliancesmanager.py
```diff
@@ -0,0 +1,148 @@
...
+    am._get_owned_appliances = get_mock_coro(False)
+
+    await am.fetch_appliances()
+
+    am._get_owned_appliances.assert_called()  # type: ignore
+    am._get_shared_appliances.assert_called()  # type: ignore
+    await http_client_mock.close_session()
+
+
+async def test_fetch_appliances_loads_lists(
```
abmantis, line 98: if we update the test above to use only the public methods, I think this can be merged into it, right?

tests/test_auth.py
```diff
@@ -7,29 +7,51 @@
...
 
+BACKEND_SELECTORS = [BackendSelectorMockMultipleCreds(), BackendSelectorMock()]
 
 pytestmark = pytest.mark.asyncio
 
 
-async def test_auth_success(http_client_mock: AiohttpClientMocker):
+@pytest.mark.parametrize("refresh_token", [None, "refresh_token_123"])
+@pytest.mark.parametrize("backend_selector", BACKEND_SELECTORS)
+async def test_get_auth_body(
```
abmantis, line 23: I think we can keep the check for the auth_data on `test_auth_success` as before, and remove this test, so that we use only public methods here too.
We simply need to construct the auth_data dict from the first element on `backend_selector.client_credentials`.
We could also add another test `test_auth_multiple_client_credentials` where the http_client_mock returns 404 for each call, and we just check that it goes trough all the `backend_selector.client_credentials`.

abmantis (changes requested):

NodeJSmith:
@abmantis  Fixed up the tests, I agree that this is a better way of doing this. Accidentally broke the PR, I attempted to rename the branch and that didn't work out well. I think it's back to normal now.

Let me know if you have any other concerns on the PR!

abmantis (approved):

abmantis:
I'm sorry for the delay in moving this forward.
Thanks a lot for your work! I will release a new version to pypi now. Let me know if you need help getting it into HA!

abmantis:
0.18.6 released

abmantis:
@NodeJSmith do you want me to update the HA component to use this?

NodeJSmith:
I've started a local branch that has the changes incorporated but I haven't gotten further than that. I'll push the branch tomorrow and start buttoning it up. Happy for any insight or contributions you want to provide though, this will be my first time contributing to HA.

abmantis:
If you just want to use the newer version of this lib I think you only need to bump the version, like this: https://github.com/home-assistant/core/pull/111439

NodeJSmith:
Unfortunately bumping it isn't enough, because the integration currently doesn't allow you to choose your brand, it chooses it for you based on region. Because the shared appliances call requires the proper brand header the integration needs to be updated to allow the user to choose their brand during configuration.

I have that done already but trying to install all of the requirements to run the tests took longer than I expected so I didn't get my changes pushed yet. I'll be doing that in the morning hopefully. I'll just need to add a few tests for the changes to the brand in the configuration and then I should be able to create a PR

PR #50, opened 2024-04-01
Add user-agent to auth header

mkmer:
Apparently whirlpool is becoming a bit more picky about the auth header - User-Agent must be "okhttp/3.12.0" or they return 403 when attempting to authenticate.

This will require are version bump so we can fix HA.

https://github.com/home-assistant/core/issues/114463


abmantis (approved):

abmantis:
@mkmer just released v0.18.7 with this fix. 
Thanks a lot! I am not able to reproduce the issue on my side, so I wouldn't be able to fix it.

PR #52, opened 2024-04-08
Cleanup - type hints, remove unused imports, remove disabled auto-renewal logic, add pre-commit, etc

NodeJSmith:
Wanted to get things cleaned up a bit before working on the EventSocket logic, to make sure that PR didn't include extraneous stuff.

Cleanup includes:

- add .flake8 file to ignore E402 (imports not at top of file, due to HA license comment in `tests/aiohttp`)
- remove comment from pyproject exclude line
- switch out `Dict`/`List` imports from `typing` with `dict` and `list`
- add type hints to all (?) code in `auth`, `appliance`, and `appliancesmanager`
- reorganized the methods in `Appliance` - this one was purely for my own benefit, I wanted to get all the ones that use `EventSocket` into one contiguous section
- remove the `__init__` methods from `WasherDryer`, `Aircon`, and `Oven` - I can add these back if necessary, but I noticed that they weren't actually doing anything different from the `Appliance` `__init__` method so it seemed like they didn't need to be there
- in the cli, dump the raw data using `json.dumps` so it's easier to read

Let me know if there are any concerns with these changes or if there are formatting conventions that this project uses that I may have missed.

Note: I didn't bump the version since there are no actual functionality changes and figured we wouldn't both bumping until there were new features/changes.


NodeJSmith:
Only set WP-CLIENT-BRAND header on shared appliances call. Bump to 0.18.8 due to that change. 

(hopefully) Fixes #https://github.com/home-assistant/core/issues/115130

@abmantis Can I get your eyes on this when you have a moment? And if you'd prefer I do the HA fix in a PR without any cleanup I can split these, just let me know

NodeJSmith:
Confirmed on the HA issue that this change does fix the issue, so we'll definitely want this in as quickly as possible

abmantis:
Thanks for this! 
Yeah, it would be great if you could split this into smaller PRs, each focused on a single change. Otherwise this will take a lot to get merged, since it will get more review rounds.

Also, it is easier to find time during the day to review small PRs :)

abmantis:
It is fine to leave the version bump out as that allows us to bundle more changes if needed, and I can do it anyway for the release

PR #55, opened 2024-04-10
add .pre-commit-config.yaml + run precommit

NodeJSmith:
Minor cleanup, just add a .pre-commit-config.yaml file and run it. Everything in the .pre-commit-config should match the standards the library was already using, including isort and black.


abmantis (approved):
Nice! I wanted to do this but never got to actually do it. Thanks again.

PR #56, opened 2024-04-10
Remove unused auto renewal logic

NodeJSmith:
Auto renewal logic has been commented out for some time, this PR removes it. It can be added back later if/when it is desired


abmantis (approved):
Thanks!

PR #58, opened 2024-04-11
Python-requires to >=3.11, type annotation updates, add pyupgrade to .pre-commit-config

### LINKED ISSUE #57: Python required version minimum
Since the main use of this library is for HA and that runs Python 3.12 as of this week, are we able to reduce the Python required version minimum? If we move up to 3.10 or 3.11 we can take advantage of some of the newer features, and I don't know if there would be any downside to it.

@abmantis Thoughts? I'm not sure if there are more users or use cases that I'm not aware of for this library

NodeJSmith:
Closes #57 

- Sets python-requires to >=3.11
- Adds pyupgrade to .pre-commit-config.yaml with --py311-plus argument
- Switches out `Dict` and `List` to builtins
- Corrects type annotation in `Appliance`
- Removes unnecessary `CredentialsDict` `TypedDict` in `backendselector.py`

No type annotations were added, just modified/corrected existing ones

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,3 +24,9 @@ repos:
     rev: 24.3.0
     hooks:
       - id: black
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.15.2
+    hooks:
+    - id: pyupgrade
+      args:
+        - --py311-plus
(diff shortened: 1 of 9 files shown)


NodeJSmith:
Lowered the requirement to 3.11, I think that's high enough to take advantage of any new features of the language while not being quite as abrupt as going straight to 3.12. Happy to switch it back to 3.12 if you think we should, but I'm good with 3.11 if you are.

abmantis (approved):
Thanks!

PR #59, opened 2024-04-11
Feature/move urls to backend selector

NodeJSmith:
- create a `types.py` module for enums
  - another branch of mine where i started work on the event socket had additional types in this file, so it felt worthwhile to go ahead and make it since it will have more stuff added to it soon
  - but if we feel like it's premature or just don't want it i'm happy to move these back to backendselector
- add url properties and url methods to backend selector
- update all use of urls to use backend_selector
- add type hints to backend selector
- remove some commented out code in conftest.py
- add a check in `test_appliancesmanager` that `WP-CLIENT-BRAND` is not in the "owned appliances" call

### DIFF
--- a/tests/test_appliancesmanager.py
+++ b/tests/test_appliancesmanager.py
@@ -29,7 +29,7 @@ def assert_appliances_manager_call(
 
     call = mock_calls[call_index]
     assert call[0] == "GET"
-    assert path.endswith(call[1].path)
+    assert BACKEND_SELECTOR_MOCK.base_url + call[1].path == path
     # call[2] is body, which will be None
 
     if required_headers is not None:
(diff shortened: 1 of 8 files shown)


tests/test_appliancesmanager.py
```diff
@@ -23,18 +23,23 @@ def assert_appliances_manager_call(
...
     path: str,
     required_headers: dict = None,
+    forbidden_headers: list = None,
 ):
     mock_calls = http_client_mock.mock_calls
 
     call = mock_calls[call_index]
     assert call[0] == "GET"
-    assert call[1].path == path
+    assert path.endswith(call[1].path)
```
abmantis, line 32: Can append `path` to `BackendSelectorMock.base_url` here, to keep the comparison stricter as it was?

abmantis (changes requested):
Nice cleaning up! Just one small comment

tests/test_appliancesmanager.py
NodeJSmith: Yep, good call, got that updated

abmantis (approved):

PR #60, opened 2024-04-12
Move to aioresponses library to simplify aiohttp tests

NodeJSmith:
@abmantis Wanted to get your thoughts on this before I start working on the other tests files. The existing fixtures and test setup has always been very confusing for me and I think aioresponses can make everything a lot simpler. 

Let me know if you have any preferences or concerns on what I have so far and I can incorporate that before moving on to the other test files.

Note: This branch is off of feature/move_urls_to_backend_selector so you'll see those commits as well until that PR is approved, it's the `test_aircon.py` file that I'm focusing on here.


NodeJSmith:
@abmantis I went ahead and finished this up - the diff makes it look like a lot of changes but is all the same tests, just using aioresponses which lets us write them a lot cleaner and clearer. I also made sure to keep all the same checks, based on your feedback on the last PR. The one exception to that is this check in test_appliancesmanager:

```python
        if account_id is None:
            m.assert_any_call(URL(BACKEND_SELECTOR_MOCK.user_details_url), "GET")
        else:
            assert (
                "GET",
                URL(BACKEND_SELECTOR_MOCK.user_details_url),
            ) not in m.requests
```

This one just checks that we do call `user_details_url` if we don't have an account id, instead of explicitly checking if it's the first call. The reason for this is that aioresponses doesn't give us a good way (that I could find at least) to determine if this was called first, as it stores the calls in a dictionary that is keyed by method and url (e.g. ("POST", "url_value.url.com")).

If you'd like me to keep looking for a way to assert that it's first let me know, I'm positive that it _can_ be done, I just didn't want to put in a ton of time to do so if we are comfortable with this level of check.

The tests `test_auth_multiple_client_credentials` and `test_auth_bad_credentials` also got slimmed down some, because we were providing data to the mock that actually wasn't required for the test assertions, so I removed that data.

The other "big" changes were moving the json data to data files instead of putting it in python, which just feels cleaner, and adding washer and dryer data files - there are no tests using them yet, but I wanted to get them in there for future tests.

NodeJSmith:
Oh and I also updated the tests GH action to test again 3.11 and 3.12 since that is what we are supporting now

NodeJSmith:
@abmantis If you need me to break this up into smaller PRs I can. I wasn't sure how to do so with this, since it's all contained to tests, but if you give me some guidance on what would be easiest to digest I can work with that. I know it's a lot to look at in one go

abmantis:
Hello!
When I started adding the tests using aiohttp I did have some trouble and the result isn't great, I agree 😅
At the time I looked at aioresponses, but for some reason I were not able to use it, and ended up doing something similar to what HA does (or did at the time). 

I haven't had the chance to review this properly yet, but from a brief look I don't think there's a need to split it. Thanks for the concern!

tests/test_aircon.py
```diff
@@ -1,228 +1,144 @@
...
-        SAID,
-        {"Sys_OpSetQuietModeEnabled": "0"},
-        CONNECT_HTTP_CALLS + 10,
-    )
-
-    await aircon.set_display_on(True)
-    assert_appliance_setter_call(
-        appliance_http_client_mock,
+async def test_attributes():
+    backend_selector_mock = BackendSelectorMock()
```
abmantis, line 32: I would prefer to keep this as a fixture and arg to the test, since that makes test dependencies explicit (same for the other tests).

tests/test_aircon.py
```diff
@@ -1,228 +1,144 @@
...
+    session = aiohttp.ClientSession()
```
abmantis, line 33: can we also make this a fixture?

tests/test_aircon.py
```diff
@@ -1,228 +1,144 @@
...
+    aircon = Aircon(
+        backend_selector_mock,
+        MagicMock(),
```
abmantis, line 36: any reason to remove the auth_mock fixture? I know it currently does not do anything, but makes this a bit more explicit too. Also, if later we need to add some mock behavior to the auth part, we just have to update the fixture method.

tests/test_aircon.py
```diff
@@ -1,228 +1,144 @@
...
+        (Aircon.set_h_louver_swing, True, {"Cavity_OpSetHorzLouverSwing": "1"}),
+        (Aircon.set_turbo_mode, False, {"Cavity_OpSetTurboMode": "0"}),
+        (Aircon.set_eco_mode, False, {"Sys_OpSetEcoModeEnabled": "0"}),
+        (Aircon.set_quiet_mode, False, {"Sys_OpSetQuietModeEnabled": "0"}),
+        (Aircon.set_display_on, True, {"Sys_DisplaySetBrightness": "4"}),
+    ],
+)
+async def test_setters(method: Callable, argument: Any, expected: dict):
+    backend_selector_mock = BackendSelectorMock()
+    expected_template = {
```
abmantis, line 105:
```suggestion
    expected_payload = {
```

tests/test_aircon.py
```diff
@@ -1,228 +1,144 @@
...
+        assert aircon.get_turbo_mode() is True
+        assert aircon.get_eco_mode() is True
+        assert aircon.get_quiet_mode() is True
+        await aircon.disconnect()
+
+    await session.close()
+
+
+@pytest.mark.parametrize(
+    ["method", "argument", "expected"],
```
abmantis, line 88:
```suggestion
    ["method", "argument", "expected_json"],
```

tests/test_aircon.py
```diff
@@ -1,228 +1,144 @@
...
+        "data": None,
+        "json": expected_template["json"],
+        "allow_redirects": True,
+        "headers": {},
+    }
+
+    with patch("whirlpool.appliance.Appliance._create_headers", return_value={}):
+        session = aiohttp.ClientSession()
+        aircon = Aircon(backend_selector_mock, MagicMock(), SAID, session)
+        url = backend_selector_mock.appliance_command_url
```
abmantis, line 124: move this down, near the line where it is first used

tests/test_aircon.py
```diff
@@ -1,228 +1,144 @@
...
+        "method": "POST",
+        "data": None,
+        "json": expected_template["json"],
+        "allow_redirects": True,
+        "headers": {},
+    }
+
+    with patch("whirlpool.appliance.Appliance._create_headers", return_value={}):
+        session = aiohttp.ClientSession()
+        aircon = Aircon(backend_selector_mock, MagicMock(), SAID, session)
```
abmantis, line 123: move this down, near the line where it is first used

tests/test_aircon.py
```diff
@@ -1,228 +1,144 @@
...
+async def test_setters(method: Callable, argument: Any, expected: dict):
+    backend_selector_mock = BackendSelectorMock()
+    expected_template = {
+        "json": {
+            "body": expected,
+            "header": {"said": SAID, "command": "setAttributes"},
+        }
+    }
+
+    call_kwargs = {
```
abmantis, line 112:
```suggestion
    post_request_call_kwargs = {
```

tests/test_aircon.py
```diff
@@ -1,228 +1,144 @@
...
-        appliance_http_client_mock,
-        SAID,
-        {"Sys_OpSetQuietModeEnabled": "0"},
-        CONNECT_HTTP_CALLS + 10,
-    )
-
-    await aircon.set_display_on(True)
-    assert_appliance_setter_call(
-        appliance_http_client_mock,
+async def test_attributes():
```
abmantis, line 31:
```suggestion
@aioresponses()
async def test_request(mocked_aioresponses):
```
I think having this as a fixture instead of the context manager makes the tests a bit cleaner, and also reduce nesting

tests/test_aircon.py
```diff
@@ -1,228 +1,144 @@
...
+        (Aircon.set_mode, Mode.Cool, {"Cavity_OpSetMode": "1"}),
+        (Aircon.set_fanspeed, FanSpeed.Auto, {"Cavity_OpSetFanSpeed": "1"}),
+        (Aircon.set_h_louver_swing, True, {"Cavity_OpSetHorzLouverSwing": "1"}),
+        (Aircon.set_turbo_mode, False, {"Cavity_OpSetTurboMode": "0"}),
+        (Aircon.set_eco_mode, False, {"Sys_OpSetEcoModeEnabled": "0"}),
+        (Aircon.set_quiet_mode, False, {"Sys_OpSetQuietModeEnabled": "0"}),
+        (Aircon.set_display_on, True, {"Sys_DisplaySetBrightness": "4"}),
+    ],
+)
+async def test_setters(method: Callable, argument: Any, expected: dict):
```
abmantis, line 103:
```suggestion
async def test_setters(method: Callable, argument: Any, expected_json: dict):
```

tests/test_aircon.py
abmantis: btw, nice idea to use parametrized tests here!

tests/test_aircon.py
abmantis: this could also be moved near its usage

abmantis (changes requested):
I have not reviewed everything yet, but have a couple of suggestions that probably also apply to the other tests.

Overall, I think this is a nice refactoring!

NodeJSmith:
@abmantis So I created some mocks to add to conftest.py and I noticed that `Auth` actually contains everything we need for all of our logic - it contains the backend selector, the auth credentials, and the session. We currently pass all of these to everything, but I think if we just accessed them from Auth it would work just as well.

Fixtures:
```python
@pytest.fixture
def aioresponses_mock():
    with aioresponses() as m:
        yield m


@pytest.fixture(params=[BackendSelectorMock, BackendSelectorMockMultipleCreds])
def backend_selector_mock(request):
    yield request.param()


@pytest_asyncio.fixture
async def auth_mock(backend_selector_mock):
    session = aiohttp.ClientSession()
    auth = Auth(backend_selector_mock, "email", "secretpass", session)
    yield auth
    await session.close()
```

Example - we previously had a backend selector mock we provided, plus the auth mock, plus the session. The new `auth_mock` fixture actually has all of these, so when we instantiate the Oven in the below test we can just use them all from the `Auth` object.

What are your thoughts on this? I don't want to pile up too many changes, but I'm not sure how to approach the tests. Does the below work for our tests or should I specifically separate out all of these pieces, even if it means more work for mocking the calls?

```python
async def test_attributes(auth_mock, aioresponses_mock):
    oven = Oven(auth_mock._backend_selector, auth_mock, SAID, auth_mock._session)
```

abmantis:
Regarding the Auth and mock thing, I would prefer to keep the fixtures instead of getting them from Auth, to keep responsibilities separated and also avoid accessing what should be internal stuff from Auth.

I am also wondering about an alternative: could/should we make Aircon/Oven/etc fixtures? That way, the tests only needs to receive that fixture as argument.

NodeJSmith:
In a general sense are you okay with all of these classes receiving instances of the same classes (backend selector, Auth, session) or should we add a review of the architecture to the backlog at some point? It seems like it could be a code smell.

I'm okay with keeping the fixtures separate, I'll break them back out.

Regarding the appliance fixtures, I think we can do that. I think they all use the same SAID, and we set the data using aioresponses, so that should be feasible. I'll take a look at it in the next day or two.

abmantis:
I actually like the fact that they are separated and each class receives the instances they need. I usually avoid classes that "agregate" things, as they tend to become dependency magnets. 
But I may be missing what you had in mind, so please do let me know if you have any thoughts on how to improve the arch.

NodeJSmith:
Whew, that took me longer than I'd planned.

I believe I implemented all of the changes you requested. I also took your advice and created mocks for the appliances, as well as mocks for the AppliancesManager and a fixture for the `aiohttp.ClientSession` with a `session` scope so we have the one for the whole duration of the tests and it closes itself. The tests look even cleaner now that they get everything through fixtures.

Two items I want to point out to you:

* I added the `with patch("whirlpool.appliance.Appliance._create_headers", return_value={}):` decorator in the appliance mocks, so that the headers were taken care of. I confirmed that all tests still pass as expected, so I think this is okay, but if you want me to do something differently here I can. My other idea was to overwrite the method with a patched method but I thought that decorating the whole fixture was probably cleaner until we have a reason to do something different.

* In test_appliancesmanager I added an autouse fixture, for the aioresponses calls that are required for all of the tests. Let me know if you have concerns with this - I wanted to make the tests look a bit cleaner but if you have concerns with an autouse fixture I can revert this. I also split out a lot of the assertions in this file into separate tests to make things more structured.

```python
@pytest.fixture(autouse=True)
def account_id_calls_fixture(aioresponses_mock, backend_selector_mock):
    """Mock the calls to get the user details and owned appliances.

    This is required by all of the tests in this file, so we use the autouse=True
    parameter to ensure that this fixture is always run.
    """
    aioresponses_mock.get(
        backend_selector_mock.user_details_url, payload={"accountId": ACCOUNT_ID}
    )
    aioresponses_mock.get(
        backend_selector_mock.get_owned_appliances_url(ACCOUNT_ID),
        payload={ACCOUNT_ID: owned_appliance_data},
    )
    aioresponses_mock.get(
        backend_selector_mock.shared_appliances_url, payload=shared_appliance_data
    )
    yield
```

tests/test_appliancesmanager.py
```diff
@@ -1,123 +1,148 @@
...
-    call = mock_calls[call_index]
-    assert call[0] == "GET"
-    assert BACKEND_SELECTOR_MOCK.base_url + call[1].path == path
-    # call[2] is body, which will be None
 
-    if required_headers is not None:
-        for k, v in required_headers.items():
-            assert call[3][k] == v
+with open(DATA_DIR / "owned_appliances.json") as f:
+    owned_appliance_data = json.load(f)
```
abmantis, line 25: since this is only used in the fixture, we can move this inside it

tests/test_appliancesmanager.py
```diff
@@ -1,123 +1,148 @@
...
 
-    if forbidden_headers is not None:
-        for k in forbidden_headers:
-            assert k not in call[3]
+with open(DATA_DIR / "shared_appliances.json") as f:
```
abmantis, line 27: same as above

tests/test_appliancesmanager.py
```diff
@@ -1,123 +1,148 @@
...
+    aioresponses_mock.get(
+        backend_selector_mock.get_owned_appliances_url(ACCOUNT_ID),
+        payload={ACCOUNT_ID: owned_appliance_data},
     )
-    mock_appliancesmanager_get_shared_appliances_get(
-        http_client_mock, BACKEND_SELECTOR_MOCK
+    aioresponses_mock.get(
+        backend_selector_mock.shared_appliances_url, payload=shared_appliance_data
     )
+    yield
```
abmantis, line 48:
```suggestion
```
is that required?

tests/test_appliancesmanager.py
```diff
@@ -1,123 +1,148 @@
...
 
-    http_client_mock.create_session(asyncio.get_event_loop())
-    auth = Auth(BACKEND_SELECTOR_MOCK, "email", "secretpass", http_client_mock.session)
 
-    if account_id is not None:
-        auth._auth_dict["accountId"] = account_id
+async def test_uer_details_not_called_if_account_id_is_known(
```
abmantis, line 51:
```suggestion
async def test_user_details_not_requested_if_account_id_is_known(
```

tests/test_appliancesmanager.py
```diff
@@ -1,123 +1,148 @@
...
+    auth_mock: Auth,
+    backend_selector_mock: BackendSelector,
+    appliances_manager: AppliancesManager,
+    aioresponses_mock,
+):
+    auth_mock._auth_dict["accountId"] = ACCOUNT_ID
```
abmantis, line 57: can we have this test without changing internal stuff of the class?
Maybe by:
- calling `fetch_appliances()` once
- checking the request was done
- resetting `aioresponses_mock.requests`
- calling `fetch_appliances()` again
- checking that the request was not done
 
If that works, we should probably rename the test to `test_user_details_requested_only_once` or something, and can remove the `test_user_details_called_if_account_id_not_known` test.

tests/test_appliancesmanager.py
```diff
@@ -1,123 +1,148 @@
...
+
+
+async def test_fetch_appliances_calls_owned_and_shared_methods(
+    appliances_manager: AppliancesManager,
+):
+    shared_appliance_mock = get_mock_coro(None)
+    owned_appliance_mock = get_mock_coro(None)
+
+    appliances_manager._get_shared_appliances = shared_appliance_mock
+    appliances_manager._get_owned_appliances = owned_appliance_mock
```
abmantis, line 128: Can we do this check using the http requests instead of altering internal behavior of the class? We could then probably also merge it `test_fetch_appliances_returns_owned_and_shared_appliances`

tests/test_appliancesmanager.py
```diff
@@ -1,123 +1,148 @@
...
-        http_client_mock, BACKEND_SELECTOR_MOCK
-    )
-
-    http_client_mock.create_session(asyncio.get_event_loop())
-    auth = Auth(BACKEND_SELECTOR_MOCK, "email", "secretpass", http_client_mock.session)
-
-    am = AppliancesManager(BACKEND_SELECTOR_MOCK, auth, http_client_mock.session)
-    am._get_shared_appliances = get_mock_coro(shared_response)
-    am._get_owned_appliances = get_mock_coro(owned_response)
+    appliances_manager._get_shared_appliances = get_mock_coro(shared_response)
```
abmantis, line 143: Could also be done using the requests instead?

abmantis (changes requested):
Tests are indeed looking good! Left just some minor comments.
I haven't reviewed them all yet, hopefully I can review all the remaining before the weekend.

tests/test_appliancesmanager.py
NodeJSmith: Good idea, I've implemented this test and removed the other one

tests/test_appliancesmanager.py
NodeJSmith: Yep, that's a good idea, I've implemented it

NodeJSmith:
@abmantis I moved `_get_account_id` from `AppliancesManager` to `Auth`, as it seems to make more sense there. The way it was previously, we could theoretically find that the account id is not in the `Auth`'s `_auth_dict` attribute, get the value using a call, and leave that empty in the `Auth` class, which doesn't make any sense.

Let me know if the changes look okay to you. I also removed the `test_fetch_appliances_returns_true_if_either_method_returns_true` test as it basically just tested boolean logic, which was silly.

I also updated the shared/owned tests to use the aioresponses assert method - this does require calling the private `_create_headers` method, but that's the only way to use the assert methods, and I think it's cleaner than the way I was doing it previously.

Oh and I fixed the cli command option, it was missing `await`s on the inputs.

Let me know if there are any concerns with the changes.

tests/test_appliancesmanager.py
```diff
@@ -45,51 +34,24 @@ def account_id_calls_fixture(aioresponses_mock, backend_selector_mock):
...
 ):
-    auth_mock._auth_dict["accountId"] = ACCOUNT_ID
-
     await appliances_manager.fetch_appliances()
 
-    assert (
-        "GET",
-        URL(backend_selector_mock.user_details_url),
-    ) not in aioresponses_mock.requests
+    headers = appliances_manager._create_headers()
```
abmantis, line 46: lets hard-code the headers here. that way we avoid accessing internal class stuff, and it also serves as a way to validate the headers and you can then remove the assert on line 54.

I don't really mind having some duplication or hardcoded strings on tests if they are used to validate the lib's behavior

whirlpool/auth.py
```diff
@@ -117,9 +117,28 @@ def is_access_token_valid(self):
     def get_access_token(self):
         return self._auth_dict.get("access_token", None)
 
-    def get_account_id(self) -> str | None:
+    async def get_account_id(self) -> str | None | bool:
```
abmantis, line 120: why not return None instead of False?

tests/test_appliancesmanager.py
```diff
@@ -99,13 +61,13 @@ async def test_get_shared_appliances_call_contains_wp_client_brand_header(
...
     await appliances_manager.fetch_appliances()
 
-    # call to get shared appliances - this one should contain the WP-CLIENT-BRAND header
-    shared_call = aioresponses_mock.requests[
-        ("GET", URL(backend_selector_mock.shared_appliances_url))
-    ]
-    assert len(shared_call) == 1
-    shared_call_headers = shared_call[0][1]["headers"]
-    assert shared_call_headers["WP-CLIENT-BRAND"] == "DUMMY_BRAND"
+    headers = appliances_manager._create_headers(include_wp_brand_name=True)
```
abmantis, line 64: same as above

tests/test_appliancesmanager.py
```diff
@@ -120,29 +82,20 @@ async def test_fetch_appliances_returns_owned_and_shared_appliances(
...
+    backend_selector_mock: BackendSelector,
+    aioresponses_mock,
 ):
-    shared_appliance_mock = get_mock_coro(None)
-    owned_appliance_mock = get_mock_coro(None)
-
-    appliances_manager._get_shared_appliances = shared_appliance_mock
-    appliances_manager._get_owned_appliances = owned_appliance_mock
+    owned_headers = appliances_manager._create_headers()
+    shared_headers = appliances_manager._create_headers(include_wp_brand_name=True)
```
abmantis, line 89: same as above

tests/test_auth.py
```diff
@@ -111,3 +114,35 @@ async def test_auth_bad_credentials(
...
+        "Cache-Control": "no-cache",
+    }
+
+    await auth_mock.get_account_id()
+
+    aioresponses_mock.assert_called_with(
+        backend_selector_mock.user_details_url, "GET", headers=headers
+    )
+
+    assert auth_mock._auth_dict["accountId"] == ACCOUNT_ID
```
abmantis, line 140:
```suggestion
```
you can reset `aioresponses_mock` and check that the call count for the URL is zero after the second call to `get_account_id()` instead of accessing the internal dict

dryer_device_data_model.json
abmantis, line 1: what are these new json files for?

tests/conftest.py
```diff
@@ -1,48 +1,70 @@
...
+
+@pytest.fixture(params=[BackendSelectorMock, BackendSelectorMockMultipleCreds])
+def backend_selector_mock(request):
+    yield request.param()
 
 
 @pytest.fixture
-def backend_selector_mock():
-    return BackendSelectorMock()
+def auth_mock(backend_selector_mock, client_session_fixture):
```
abmantis, line 40:
```suggestion
pytest.fixture(name="auth")
def auth_fixture(backend_selector_mock, client_session_fixture):
```
seeing "auth_mock" in the tests where Auth is being tested was getting me confused, since it is not a mock

tests/conftest.py
```diff
@@ -1,48 +1,70 @@
...
+    auth = Auth(backend_selector_mock, "email", "secretpass", client_session_fixture)
+    yield auth
 
 
-@pytest.fixture
-def auth_mock():
-    return MagicMock()
+@pytest.fixture(name="oven")
+def oven_mock(backend_selector_mock, auth_mock, client_session_fixture):
```
abmantis, line 46:
```suggestion
def oven_fixture(backend_selector_mock, auth_mock, client_session_fixture):
```

tests/conftest.py
```diff
@@ -1,48 +1,70 @@
...
+    with patch("whirlpool.appliance.Appliance._create_headers", return_value={}):
+        oven = Oven(backend_selector_mock, auth_mock, SAID, client_session_fixture)
+        yield oven
+
+
+@pytest.fixture(name="washer_dryer")
+def washer_dryer_mock(backend_selector_mock, auth_mock, client_session_fixture):
```
abmantis, line 53:
```suggestion
def washer_dryer_fixture(backend_selector_mock, auth_mock, client_session_fixture):
```

tests/conftest.py
```diff
@@ -1,48 +1,70 @@
...
+    with patch("whirlpool.appliance.Appliance._create_headers", return_value={}):
+        washer_dryer = WasherDryer(
+            backend_selector_mock, auth_mock, SAID, client_session_fixture
+        )
+        yield washer_dryer
+
+
+@pytest.fixture(name="aircon")
+def aircon_mock(backend_selector_mock, auth_mock, client_session_fixture):
```
abmantis, line 62:
```suggestion
def aircon_fixture(backend_selector_mock, auth_mock, client_session_fixture):
```

tests/conftest.py
```diff
@@ -1,48 +1,70 @@
...
+        yield aircon
 
 
-@pytest.fixture(autouse=True)
-def event_socket_mock(mocker):
-    event_socket = mocker.patch("whirlpool.appliance.EventSocket").return_value
-    event_socket.stop = AsyncMock()
-    return event_socket
+@pytest.fixture(name="appliances_manager")
+def appliances_manager_mock(backend_selector_mock, auth_mock, client_session_fixture):
```
abmantis, line 69:
```suggestion
def appliances_manager_fixture(backend_selector_mock, auth_mock, client_session_fixture):
```

tests/test_auth.py
```diff
@@ -43,78 +39,110 @@ async def test_auth_success(http_client_mock: AiohttpClientMocker):
...
-        "jti": "?????",
-    }
+    # don't need repeat here because the first one will succeed
+    # so we will only call this url once
+    aioresponses_mock.post(AUTH_URL, payload=mock_resp_data)
+
+    await auth_mock.do_auth(store=False)
+    assert auth_mock.is_access_token_valid()
+    assert auth_mock.get_said_list() == ["SAID1", "SAID2"]
+
```
abmantis, line 50:
```suggestion
    assert auth_mock.get_account_id() == ACCOUNT_ID
    
```

tests/test_appliancesmanager.py
```diff
@@ -1,123 +1,101 @@
...
 
-    if required_headers is not None:
-        for k, v in required_headers.items():
-            assert call[3][k] == v
 
-    if forbidden_headers is not None:
-        for k in forbidden_headers:
-            assert k not in call[3]
+@pytest.fixture(autouse=True)
+def account_id_calls_fixture(aioresponses_mock, backend_selector_mock):
```
abmantis, line 15:
```suggestion
def common_http_calls_fixture(aioresponses_mock, backend_selector_mock):
```
I was going to suggest `user_details_and_appliances_calls_fixture`, but it was a bit too big...

tests/test_appliancesmanager.py
abmantis: if we use hardcoded headers in this test, then do we need the `test_get_owned_appliances_call_does_not_contain_wp_client_brand_header` and `test_get_shared_appliances_call_contains_wp_client_brand_header` tests? maybe we can remove them

tests/test_appliancesmanager.py
```diff
@@ -1,123 +1,101 @@
...
+        backend_selector_mock.get_owned_appliances_url(ACCOUNT_ID),
+        "GET",
+        headers=owned_headers,
+    )
 
-    assert result == bool(owned_response or shared_response)
-    await http_client_mock.close_session()
+    aioresponses_mock.assert_called_with(
+        backend_selector_mock.shared_appliances_url, "GET", headers=shared_headers
+    )
```
abmantis, line 101: can we also reset `aioresponses_mock`, call `fetch_appliances` again and assert that the call count to `aioresponses_mock` is zero?

abmantis (changes requested):

tests/test_appliancesmanager.py
NodeJSmith: Yep, good call, I removed those after hardcoding the headers

tests/test_appliancesmanager.py
NodeJSmith: I didn't follow this comment, why would we do this? fetch_appliances will be calling the endpoints every time, it doesn't cache return appliances

NodeJSmith:
@abmantis Sorry to leave you hanging for a month, finally got around to implementing your requested changes, except for the last comment that I've left unresolved, I didn't follow what you were looking for on that one.

tests/test_appliancesmanager.py
abmantis: Well, I don't remember why I suggested that. Now it does not seem to make sense at all :rofl:

abmantis:
Thanks for getting back at this @NodeJSmith ! Everything LGTM. Are the 4 new json files needed, or can they be removed?

NodeJSmith:
I don't think the paths files or the data model files are required, I think those got added accidentally when I'd been messing with something else. I'll remove them

abmantis (approved):
Again, thanks a lot for this!

PR #63, opened 2024-12-13
Consul beer fridges integration

leandrocaldas:
hi @abmantis  I inserted a new integration equipemt smartbeer, it uses a US whirlpool api, it is a Brazilian brand called consul, it is a fridge dedicated to beer, and there are some functionality, I tryed following the current structure but if has some thing to fix let me know. Now I should develop the Home assistant side, but I never delevoled some thing to HA.

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -22,7 +22,7 @@ parser = argparse.ArgumentParser()
 parser.add_argument("-e", "--email", help="Email address")
 parser.add_argument("-p", "--password", help="Password")
 parser.add_argument(
-    "-b", "--brand", help="Brand (whirlpool/maytag/kitchenaid)", default="whirlpool"
+    "-b", "--brand", help="Brand (whirlpool/maytag/kitchenaid/consul)", default="whirlpool"
 )
 parser.add_argument("-r", "--region", help="Region (EU/US)", default="EU")
 parser.add_argument("-l", "--list", help="List appliances", action="store_true")
@@ -40,6 +40,8 @@ async def start():
         selected_brand = Brand.Maytag
     elif args.brand == "kitchenaid":
         selected_brand = Brand.KitchenAid
+    elif args.brand == "consul":
+        selected_brand = Brand.Consul
     else:
         logger.error("Invalid brand argument")
         return
@@ -66,6 +68,7 @@ async def start():
             print(appliance_manager.aircons)
             print(appliance_manager.washer_dryers)
             print(appliance_manager.ovens)
+            print(appliance_manager.beer_fridges)
             return
 
         if not args.said:
--- a/cli.py
+++ b/cli.py
@@ -5,7 +5,7 @@ import logging
 import aiohttp
 
 from cli_ac_menu import show_aircon_menu
-from cli_beer_fridge_menu import show_beer_fridge_menu
+from cli_refrigerator_menu import show_refrigerator_menu
 from cli_oven_menu import show_oven_menu
 from cli_washerdryer_menu import show_washerdryer_menu
 from whirlpool.appliancesmanager import AppliancesManager
@@ -72,8 +72,8 @@ async def start():
             print(appliance_manager.aircons)
             print(appliance_manager.washer_dryers)
             print(appliance_manager.ovens)
-            print(appliance_manager.beer_fridges)
-            # return
+            print(appliance_manager.refrigerator)
+            return
 
         for ac_data in appliance_manager.aircons:
             if ac_data["SAID"] == args.said:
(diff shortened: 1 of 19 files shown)


cli.py
```diff
@@ -87,5 +90,16 @@ def attr_upd():
                 await show_oven_menu(backend_selector, auth, args.said, session)
                 return
 
+        for bf_data in appliance_manager.beer_fridges:
+            if bf_data["SAID"]:
```
abmantis, line 94: This would always select the first fridge, even if the user has more than 1.
any reason to deviate from the way the other appliances' selection works? I would prefer to keep it consistent (passing the SAID as an argument)

cli.py
```diff
@@ -66,11 +72,8 @@ def attr_upd():
             print(appliance_manager.aircons)
             print(appliance_manager.washer_dryers)
             print(appliance_manager.ovens)
-            return
-
-        if not args.said:
-            logger.error("No appliance specified")
-            return
+            print(appliance_manager.beer_fridges)
+            # return
```
abmantis, line 76:
```suggestion
            return
```
the `--list` should only show the list output and terminate

whirlpool/beer_fridge.py
```diff
@@ -0,0 +1,56 @@
...
+            await self.send_attributes({SETTING_TEMP: str(temp)})
+
+    async def set_especific_temp(self, temp: int):
+        allowed_temps = [-4, -2, 0, 3, 5]
+        if temp not in allowed_temps:
+            raise ValueError(
+                f"Invalid temperature: {temp}. Allowed values are {allowed_temps}."
+            )
+        await self.send_attributes({SETTING_TEMP: str(TEMP_MAP[temp])})
+
```
abmantis, line 43:
```suggestion
```
Since we have `set_temp` already, can we remove this one since it duplicates behavior?

abmantis (changes requested):
Thanks for the contribution! Left a couple of suggestions.

It seems that the options are not beer specific and could work with any smart fridge and not only for bear fridges. 
Should we rename things to `refrigerator` instead of `beer_fridge`? It seems that `refrigerator` is the term used by whirlpool anyway

Also, can you please add tests for the new appliance? Thanks

cli.py
leandrocaldas: I used it to reeuse list result

leandrocaldas:
Thanks for your review, I updated the PR, let me know if you have any other concern about this implementation.

abmantis (changes requested):

cli.py
```diff
@@ -87,5 +94,14 @@ def attr_upd():
...
                 return
 
+        for rf_data in appliance_manager.refrigerators:
+            if rf_data["SAID"] == args.said:
+                await show_refrigerator_menu(backend_selector, auth, args.said, session)
+                return
+
+        if not args.said:
+            logger.error("No appliance specified")
+            return
```
abmantis, line 104:
```suggestion
```
this check is already done above

cli_refrigerator_menu.py
```diff
@@ -0,0 +1,72 @@
...
+    loop = True
+    while loop:
+        print_menu()
+        choice = await aioconsole.ainput("Enter your choice: ")
+
+        if choice == "+":
+            temp = rf.get_temp() - 1
+            await rf.set_temp(temp)
+        elif choice == "-":
+            temp = rf.get_temp() + 1
```
abmantis, line 47:
```suggestion
        if choice == "-":
            temp = rf.get_temp() - 1
            await rf.set_temp(temp)
        elif choice == "+":
            temp = rf.get_temp() + 1
```

abmantis (changes requested):
Just two small changes, otherwise LGTM!

cli_refrigerator_menu.py
leandrocaldas: It should be kept this way because the higher the number, the colder it will be, as you see:
TEMP_MAP = {
    -4: 12,
    -2: 11,
    0: 10,
    3: 9,
    5: 8,
}

abmantis (approved):

abmantis:
Thanks a lot!

leandrocaldas:
Very nice, now could you include this integration in HA? Any question let me know and I can test it during development.

abmantis:
Hey. I can't promise, but I'll add it to my TODO list. 
If you do want to contribute to HA I would also be happy to help!

lccmachado:
Hi @abmantis was this feature inserted in .19 release ?
Thx for all contributions!

abmantis:
@lccmachado yes. It was already in 0.18.12.
But someone still needs to add support to the HA integration. If you want to give it a shot and need help, feel free to ping me on Discord!

abmantis:
Hey @leandrocaldas !
Have you had the chance to test it out? Since I don't have a way to test it, I'm keeping the PR in draft for now.

PR #74, opened 2025-01-09
Device registry

lllucius:
This converts all of the Appliance creation to a "registry" of sorts. So when a new device type is added, say a humidifer, then you just need to create the humidifer.py and import it in __init__.py


abmantis:
Marking this as draft too, since it as a lot of unrelated changes from the other branches (those should get merged in first, before this one can go)

lllucius:
It ws more like me just being stupid for skipping running pytest in the first.  I'd never use it in the past, but for this kind of project I definitely see the benefit.

I see the Ruff is complaining.  Did you want me to address those failures as well?

lllucius:
Man, I'm liking this Ruff fella!  I try to be consistent with "style" across files, but I'm always missing something.  Ruff is a miracle worker!!!

abmantis:
Sorry, but if you look at the diff you can see that this has a lot of changes from the websocket move PR for example.

Also, I am not sure I understand the motivation for this PR, but since it contains a lot of changes I haven't looked too deep into it.

lllucius:
The reason I carried all of the changes from the previous PRs is be cause the successive fellas depeneded on those changes.

 �

The purpose of this PR was to move the device type determination out of AppliancesManager and into the control of the actual device classes.   So, one benefit is that when new devices are added, AppliancesManager doesn’t have to be modified…just __init__.py, types.py (to add a new ApplianceKind), and the device code itself.

 �

lllucius:
Here’s a comparison between this PR branch and the previous PR branch:

 �

https://github.com/lllucius/whirlpool-sixth-sense/compare/washer_dryer_split...lllucius:whirlpool-sixth-sense:device_registry

 �

lllucius:
Getting a bit difficult to maintain here and would rather just resubmit a clean PR after the other 2 have been resolved.

PR #119, opened 2026-01-28
Implement periodic keepalive for oven websocket

### LINKED ISSUE #116: Possible cause for oven data timeouts
When monitoring oven state in the latest version of the HA Whirlpool integration, the websocket data seems to eventually time out, and no further updates are received until the integration is reloaded.  This occurs despite the fact that washers and dryers continue use the same methods to update their data, and my own washer and dryer continue to update state after the oven no longer does.

While I haven't found a specific code issue, I discovered that if I open the KitchenAid app for my oven, the state updates immediate resume again in Home Assistant, without the need to reload the integration.

I decided to set up a proxy capture for the KitchenAid app to see what might be going on behind the scenes.  This particular POST call to `https://api.whrcloud.com/api/v3/registerDeviceIDForPush` stood out to me:

Image

and it's response:

Image

I haven't proven that this is the "magic call", but I suspect it is.  I don't see it referenced in `whirlpool-sixth-sense` at this time.  It probably needs to be called periodically to maintain the web socket connection.

I'll share more if I learn more, but wanted to share the observation.

MizterB:
Resolves https://github.com/abmantis/whirlpool-sixth-sense/issues/116

Implements a call to `Oven.fetch_data()` every 20 minutes if there is at least 1 oven defined in the AppliancesManager.  This prevents the oven websocket connection from going stale.

### DIFF
--- a/tests/test_oven.py
+++ b/tests/test_oven.py
@@ -1,9 +1,12 @@
+import asyncio
 from collections.abc import Callable
+from typing import cast
 
 import pytest
 from aioresponses import aioresponses
 from yarl import URL
 
+import whirlpool.oven as oven_module
 from whirlpool.appliancesmanager import AppliancesManager
 from whirlpool.auth import Auth
 from whirlpool.backendselector import BackendSelector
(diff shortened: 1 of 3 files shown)


MizterB:
@abmantis Wanted to let you  know that I've been using this approach for several months now, and it is working without issue.  No rate limits being hit.

abmantis:
Hey! Sorry for the time getting back to this!

I've recalled that this was added a few years ago, to solve a similar (or the same?) issue: https://github.com/home-assistant/core/pull/97222 (see the issue linked in the PR). So for people with washer/dryers, it already polls every 5min, which might explain why we haven't heard more people complaining.

I think we can simplify the implementation, and make it general (move it out of the oven). So, my suggestion would be to implement it in the `ApplianceManager`, pick one appliance from `self.all_appliances`, and do the periodic `fetch_data()` there. Does that make sense?

After merging this we can probably remove the polling from Home Assistant, since that is less efficient (polls multiple appliances).

abmantis:
Closing in favor of https://github.com/abmantis/whirlpool-sixth-sense/pull/131

PR #124, opened 2026-04-13
feat(awsiot): microwave support + factory pattern + capability download

pts211:
## Summary

- **Factory pattern** (`factory.py`, `matchers.py`): Priority-based `@register_appliance` decorator routes IoT things to concrete appliance classes by category, addressees, features, or model prefix
- **Capability download** (`capabilities.py`): `CapabilityDownloader` fetches capability XML via MQTT, parses into `CapabilityProfile` (addressees, features, commands, metadata), with memory + disk cache
- **Microwave ABC** (`whirlpool/microwave.py`): Full abstract base with enums for cavity state, door status, hood fan/light, cook lifecycle, recipe IDs
- **Concrete AWS microwave** (`awsiot/microwave.py`): Translates MQTT state paths to the ABC — cavity state, door, cook timer, hood fan/light/color, control lock, quiet/sabbath modes, start/cancel cook with remote-start gate
- **Slim `Appliance` base** (`awsiot/appliance.py`): Deep-merge state deltas, `_get_path_*` helpers, presence detection with auto-refetch on reconnect
- **Refactored `AppliancesManager`** (`awsiot/appliancesmanager.py`): Uses factory + capability download instead of hardcoded routing
- **Fixture capture tool** (`tools/capture_fixtures.py`): `--all` captures every appliance on an account, `--redact` scrubs SAIDs/serials, filenames use `{model}-{said_hash}` to avoid collisions across same-category devices
- **Real device fixtures**: Captured from a KitchenAid KMML550RPS00 microwave (anonymized), including full capability XML and state snapshots
- **72 awsiot tests**: Unit tests for all new modules, integration test for full discovery → getters → setters → state delta chain

Relates to #117 and #122.

## Scope

This PR intentionally ships only the AWS IoT **scaffolding** + **microwave** as the reference implementation. Additional appliance classes on the AWS path (laundry, oven, refrigerator, cooktop, dishwasher) will follow in separate PRs once capability data is captured from real devices — the factory + capability-download plumbing here is designed to accommodate them without further refactors.

## Test plan

- [x] All 72 `tests/awsiot/` tests pass (`pytest tests/awsiot/`)
- [x] Ruff clean (`ruff check .`)
- [x] Real-device validation on a KitchenAid KMML550RPS00: discovery, all getters, all setters (cavity light, hood fan/light/color, quiet mode, control lock, sabbath), cook lifecycle (start/cancel), door open/close, presence detection (unplug/replug)
[...]

PR #126, opened 2026-04-14
AWS IoT (1/4): Microwave appliance support

pts211:
## Summary

Routes the `cooking` AWS IoT thing category to a new `Microwave` class instead of `Oven`, and exposes read-only getters plus a single `set_cavity_light` setter.

This is PR 1 of 4 splitting out #124 per review feedback. Keeping this minimal: only the wire-up and the getters/setter that work unconditionally on every microwave. Other setters (hood fan/light, quiet/sabbath/control lock) are intentionally deferred to a later PR — those require a capability profile check and would send commands to appliances that don't support them.

## Changes

- `whirlpool/microwave.py` — transport-agnostic `Microwave` ABC with `MicrowaveCavityState`, `MicrowaveDoorStatus`, `HoodFanSpeed`, `HoodLightLevel`, `HoodLightColor` enums
- `whirlpool/awsiot/microwave.py` — concrete AWS IoT implementation on top of `awsiot.Appliance`
- `whirlpool/awsiot/appliancesmanager.py` — `cooking` category now constructs `Microwave` instead of `Oven`, adds `microwaves` accessor
- `whirlpool/appliancesmanager.py` — facade forwards `microwaves` from the AWS IoT backend (no HTTP microwaves exist)
- `tests/test_microwave.py` — 8 integration tests using a fake `MqttClient` and fake `Things` API (no `_state` poking)
- `tests/conftest.py` — fixture rewritten to call `.connect()` directly (was calling the removed `fetch_appliances()` / `all_appliances` on the facade, which had CI red)
- `whirlpool/awsiot/appliance.py` — wrap the commented-out capability-download TODO onto shorter lines (ruff E501)

## Test plan
- [x] `ruff check` — clean
- [x] `basedpyright` — 0 errors
- [x] `pytest tests/test_microwave.py` — 8 passed

Note on pytest CI: the 48 pre-existing errors in `test_aircon.py` / `test_oven.py` / `test_washer.py` / etc. are unrelated to this PR. They hit `ClientConnectionError` on unmocked `/oauth/token` because their test data was never updated for the post-split fixture API. Same failures show on #124 pytest. Out of scope here.

## Follow-ups (draft PRs to come)
- PR 2: presence/online state via `$aws/events/presence/*` topics
- PR 3: capability profile wiring + capability-gated setters (hood, quiet, sabbath, control lock)
- PR 4 (#125): `tools/capture_fixtures.py`

### DIFF
--- a/whirlpool/awsiot/microwave.py
+++ b/whirlpool/awsiot/microwave.py
@@ -1,9 +1,6 @@
 """Concrete awsiot Microwave — translates MQTT state to the Microwave ABC."""
 
-from __future__ import annotations
-
-import time
-from typing import Any, override
+from typing import override
 
 from ..microwave import (
     HoodFanSpeed,
@@ -30,58 +27,36 @@ _DOOR_STATUS_MAP: dict[str, MicrowaveDoorStatus] = {
     "closed": MicrowaveDoorStatus.Closed,
 }
 
-_HOOD_FAN_MAP: dict[str, HoodFanSpeed] = {e.value: e for e in HoodFanSpeed}
-_HOOD_LIGHT_MAP: dict[str, HoodLightLevel] = {e.value: e for e in HoodLightLevel}
+_HOOD_FAN_MAP: dict[str, HoodFanSpeed] = {
+    "off": HoodFanSpeed.Off,
+    "low": HoodFanSpeed.Low,
+    "med": HoodFanSpeed.Medium,
+    "high": HoodFanSpeed.High,
+    "boost": HoodFanSpeed.Boost,
+}
+_HOOD_LIGHT_MAP: dict[str, HoodLightLevel] = {
+    "off": HoodLightLevel.Off,
+    "low": HoodLightLevel.Low,
+    "med": HoodLightLevel.Medium,
+    "high": HoodLightLevel.High,
+}
 _HOOD_LIGHT_COLOR_MAP: dict[str, HoodLightColor] = {
-    e.value: e for e in HoodLightColor
+    "warmWhite": HoodLightColor.WarmWhite,
+    "naturalWhite": HoodLightColor.NaturalWhite,
+    "coolWhite": HoodLightColor.CoolWhite,
 }
 
 
 class Microwave(MicrowaveABC, Appliance):
-    def _get_path(self, *path: str) -> Any:
-        """Walk the state dict along `path`; return None if any step is missing."""
-        value: Any = self._data_dict
-        for key in path:
-            if not isinstance(value, dict):
-                return None
-            value = value.get(key)
-            if value is None:
-                return None
-        return value
-
-    def _get_path_str(self, *path: str) -> str | None:
-        value = self._get_path(*path)
-        return value if isinstance(value, str) else None
-
-    def _get_path_bool(self, *path: str) -> bool | None:
-        value = self._get_path(*path)
-        return value if isinstance(value, bool) else None
-
-    def _get_path_int(self, *path: str) -> int | None:
-        value = self._get_path(*path)
-        if isinstance(value, bool):
-            return None
-        return int(value) if isinstance(value, (int, float)) else None
-
-    def _get_path_float(self, *path: str) -> float | None:
-        value = self._get_path(*path)
-        if isinstance(value, bool):
-            return None
-        return float(value) if isinstance(value, (int, float)) else None
-
     @override
-    def get_cavity_state(self) -> MicrowaveCavityState:
+    def get_cavity_state(self) -> MicrowaveCavityState | None:
         raw = self._get_path_str("primaryCavity", "cavityState")
-        if raw is None:
-            return MicrowaveCavityState.Unknown
-        return _CAVITY_STATE_MAP.get(raw, MicrowaveCavityState.Unknown)
+        return _CAVITY_STATE_MAP.get(raw) if raw is not None else None
 
     @override
-    def get_door_status(self) -> MicrowaveDoorStatus:
+    def get_door_status(self) -> MicrowaveDoorStatus | None:
         raw = self._get_path_str("primaryCavity", "doorStatus")
-        if raw is None:
-            return MicrowaveDoorStatus.Unknown
-        return _DOOR_STATUS_MAP.get(raw, MicrowaveDoorStatus.Unknown)
+        return _DOOR_STATUS_MAP.get(raw) if raw is not None else None
 
     @override
     def get_door_locked(self) -> bool | None:
@@ -89,7 +64,10 @@ class Microwave(MicrowaveABC, Appliance):
         if isinstance(value, bool):
             return value
         if isinstance(value, str):
-            return value == "locked"
+            if value == "locked":
+                return True
+            if value == "unlocked":
+                return False
         return None
 
     @override
@@ -112,14 +90,21 @@ class Microwave(MicrowaveABC, Appliance):
         raw = self._get_path_str("temperatureUnit")
         if raw is None:
             return None
-        return "F" if raw.lower().startswith("f") else "C"
+        normalized = raw.strip().lower()
+        if normalized in {"f", "fahrenheit"}:
+            return "F"
+        if normalized in {"c", "celsius"}:
+            return "C"
+        return None
 
     @override
     def get_turntable_enabled(self) -> bool | None:
         raw = self._get_path_str("primaryCavity", "turnTable")
-        if raw is None or raw == "":
-            return None
-        return raw == "on"
+        if raw == "on":
+            return True
+        if raw == "off":
+            return False
+        return None
 
     @override
     def get_active_recipe_id(self) -> str | None:
@@ -143,28 +128,23 @@ class Microwave(MicrowaveABC, Appliance):
         return self._get_path_int("primaryCavity", "cookTimer", "time")
 
     @override
-    def get_cook_timer_remaining_seconds(self) -> int | None:
-        time_complete = self._get_path_int(
-            "primaryCavity", "cookTimer", "timeComplete"
-        )
-        if time_complete is not None:
-            return max(0, time_complete - int(time.time()))
-        return self.get_cook_timer_total_seconds()
+    def get_cook_timer_time_complete(self) -> int | None:
+        return self._get_path_int("primaryCavity", "cookTimer", "timeComplete")
 
     @override
     def get_hood_fan_speed(self) -> HoodFanSpeed | None:
         raw = self._get_path_str("hoodFan", "userFanSpeed")
-        return _HOOD_FAN_MAP.get(raw) if raw else None
+        return _HOOD_FAN_MAP.get(raw) if raw is not None else None
 
     @override
     def get_hood_light_level(self) -> HoodLightLevel | None:
         raw = self._get_path_str("hoodLight")
-        return _HOOD_LIGHT_MAP.get(raw) if raw else None
+        return _HOOD_LIGHT_MAP.get(raw) if raw is not None else None
 
     @override
     def get_hood_light_color(self) -> HoodLightColor | None:
         raw = self._get_path_str("hoodLightColor")
-        return _HOOD_LIGHT_COLOR_MAP.get(raw) if raw else None
+        return _HOOD_LIGHT_COLOR_MAP.get(raw) if raw is not None else None
 
     @override
     def get_remote_start_enabled(self) -> bool | None:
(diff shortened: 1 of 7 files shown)


whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
+"""Concrete awsiot Microwave — translates MQTT state to the Microwave ABC."""
+
+from __future__ import annotations
+
+import time
+from typing import Any, override
```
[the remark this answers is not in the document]
abmantis: This is fine. I'll bump it to 3.14 in pyproject

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
+"""Concrete awsiot Microwave — translates MQTT state to the Microwave ABC."""
+
+from __future__ import annotations
```
abmantis:
```suggestion
```
do we need this?

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
...
+_DOOR_STATUS_MAP: dict[str, MicrowaveDoorStatus] = {
+    "open": MicrowaveDoorStatus.Open,
+    "closed": MicrowaveDoorStatus.Closed,
+}
+
+_HOOD_FAN_MAP: dict[str, HoodFanSpeed] = {e.value: e for e in HoodFanSpeed}
+_HOOD_LIGHT_MAP: dict[str, HoodLightLevel] = {e.value: e for e in HoodLightLevel}
+_HOOD_LIGHT_COLOR_MAP: dict[str, HoodLightColor] = {
+    e.value: e for e in HoodLightColor
+}
```
abmantis, line 47: lets define the maps literally/explicitly here, instead of dynamically constructing them for HoodFanSpeed and the other enums, so that the aws microwaves are detached from whatever is specified in the enums (which can change when we add non-aws microwaves)

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
...
+        return value if isinstance(value, str) else None
+
+    def _get_path_bool(self, *path: str) -> bool | None:
+        value = self._get_path(*path)
+        return value if isinstance(value, bool) else None
+
+    def _get_path_int(self, *path: str) -> int | None:
+        value = self._get_path(*path)
+        if isinstance(value, bool):
+            return None
```
abmantis:
```suggestion
```
already handled bellow

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
...
+
+_HOOD_FAN_MAP: dict[str, HoodFanSpeed] = {e.value: e for e in HoodFanSpeed}
+_HOOD_LIGHT_MAP: dict[str, HoodLightLevel] = {e.value: e for e in HoodLightLevel}
+_HOOD_LIGHT_COLOR_MAP: dict[str, HoodLightColor] = {
+    e.value: e for e in HoodLightColor
+}
+
+
+class Microwave(MicrowaveABC, Appliance):
+    def _get_path(self, *path: str) -> Any:
```
abmantis: we can move this and the other `_get_path_*` methods to the `awsiot.Appliance` class already

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
...
+        value = self._get_path(*path)
+        if isinstance(value, bool):
+            return None
+        return float(value) if isinstance(value, (int, float)) else None
+
+    @override
+    def get_cavity_state(self) -> MicrowaveCavityState:
+        raw = self._get_path_str("primaryCavity", "cavityState")
+        if raw is None:
+            return MicrowaveCavityState.Unknown
```
abmantis: lets remove the `Unknown` state from the enum and return `None`. it aligns with the other methods. 
same for the other enums used here

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
...
+            return MicrowaveDoorStatus.Unknown
+        return _DOOR_STATUS_MAP.get(raw, MicrowaveDoorStatus.Unknown)
+
+    @override
+    def get_door_locked(self) -> bool | None:
+        value = self._get_path("primaryCavity", "doorLockStatus")
+        if isinstance(value, bool):
+            return value
+        if isinstance(value, str):
+            return value == "locked"
```
abmantis: have you seen the value being both a bool and a str?

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
...
+    @override
+    def get_cavity_light(self) -> bool | None:
+        return self._get_path_bool("primaryCavity", "cavityLight")
+
+    @override
+    async def set_cavity_light(self, on: bool) -> bool:
+        self._send_command(
+            "set", {"addressee": "primaryCavity", "cavityLight": on}
+        )
+        return True
```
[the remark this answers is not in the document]
abmantis: we should eventually improve `_send_commant` to check for success, but no need for this PR

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
...
+    def get_display_temperature(self) -> float | None:
+        return self._get_path_float("primaryCavity", "ovenDisplayTemperature")
+
+    @override
+    def get_display_temperature_unit(self) -> str | None:
+        raw = self._get_path_str("temperatureUnit")
+        if raw is None:
+            return None
+        return "F" if raw.lower().startswith("f") else "C"
+
```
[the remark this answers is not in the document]
abmantis: Lets address this one

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
...
+    @override
+    def get_turntable_enabled(self) -> bool | None:
+        raw = self._get_path_str("primaryCavity", "turnTable")
+        if raw is None or raw == "":
+            return None
+        return raw == "on"
```
abmantis:
```suggestion
        if raw == "on":
            return True
        elif raw == "off":
            return False
        else:
            return None
```

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
...
+        return self._get_path_str("primaryCavity", "cookTimer", "state")
+
+    @override
+    def get_cook_timer_total_seconds(self) -> int | None:
+        return self._get_path_int("primaryCavity", "cookTimer", "time")
+
+    @override
+    def get_cook_timer_remaining_seconds(self) -> int | None:
+        time_complete = self._get_path_int(
+            "primaryCavity", "cookTimer", "timeComplete"
```
abmantis: if timeComplete contains an unix timestamp, lets return that. Makes it easier on the HA side to use, since if it was a "count" we would have to revert it back to a timestamp there.
Please also rename the method to match.

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
...
+        )
+        if time_complete is not None:
+            return max(0, time_complete - int(time.time()))
+        return self.get_cook_timer_total_seconds()
```
abmantis: if the data is missing, we should return None

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
...
+
+    @override
+    def get_hood_fan_speed(self) -> HoodFanSpeed | None:
+        raw = self._get_path_str("hoodFan", "userFanSpeed")
+        return _HOOD_FAN_MAP.get(raw) if raw else None
```
abmantis:
```suggestion
        return _HOOD_FAN_MAP.get(raw)
```
doesn't this work? I think `get(None)` will just return `None` since its not one the map. or does the type checker complain?

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
...
+
+    @override
+    def get_hood_light_level(self) -> HoodLightLevel | None:
+        raw = self._get_path_str("hoodLight")
+        return _HOOD_LIGHT_MAP.get(raw) if raw else None
```
abmantis:
```suggestion
        return _HOOD_LIGHT_MAP.get(raw)
```

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
...
+
+    @override
+    def get_hood_light_color(self) -> HoodLightColor | None:
+        raw = self._get_path_str("hoodLightColor")
+        return _HOOD_LIGHT_COLOR_MAP.get(raw) if raw else None
```
abmantis:
```suggestion
        return _HOOD_LIGHT_COLOR_MAP.get(raw)
```

whirlpool/awsiot/microwave.py
```diff
@@ -0,0 +1,183 @@
...
+        raw = self._get_path_str("hoodFan", "userFanSpeed")
+        return _HOOD_FAN_MAP.get(raw) if raw else None
+
+    @override
+    def get_hood_light_level(self) -> HoodLightLevel | None:
+        raw = self._get_path_str("hoodLight")
+        return _HOOD_LIGHT_MAP.get(raw) if raw else None
+
+    @override
+    def get_hood_light_color(self) -> HoodLightColor | None:
```
abmantis, line 145: they even have colors for the light?! 😆

abmantis (changes requested):
Still have to review the tests, both overall this looks quite good.
Left a few minor suggestions.

whirlpool/awsiot/microwave.py
pts211: Done in 4fd0fde. The AWS microwave maps are now defined explicitly so they are decoupled from the shared enums.

whirlpool/awsiot/microwave.py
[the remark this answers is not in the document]
pts211: Addressed in 4fd0fde. It now only accepts F/fahrenheit and C/celsius, and returns None otherwise.

whirlpool/awsiot/microwave.py
pts211: Didn't need it. Removed in 4fd0fde.

whirlpool/awsiot/microwave.py
pts211: I kept that guard. bool is a subclass of int, so otherwise True/False would turn into 1/0.

whirlpool/awsiot/microwave.py
pts211: Done in 4fd0fde. Moved the _get_path_* helpers into awsiot.Appliance.

whirlpool/awsiot/microwave.py
pts211: Done in 4fd0fde. Removed the Unknown enum values and return None instead.

whirlpool/awsiot/microwave.py
pts211: Yes. The fixture here has a string value, and the Android parser treats it as a bool, so I kept both and narrowed the string handling.

whirlpool/awsiot/microwave.py
pts211: Done in 4fd0fde. It now maps "on" to True, "off" to False, and returns None otherwise.

whirlpool/awsiot/microwave.py
pts211: Done in 4fd0fde. It now returns the raw timeComplete timestamp via get_cook_timer_time_complete().

whirlpool/awsiot/microwave.py
pts211: Yep, it returns None when the data is missing now.

whirlpool/awsiot/microwave.py
pts211: Yes at runtime, but raw is str | None here and the type checker complains, so I kept the explicit None guard.

whirlpool/awsiot/microwave.py
pts211: Same reason here, kept the None guard for typing.

whirlpool/awsiot/microwave.py
pts211: Same here, kept the None guard for typing.

whirlpool/awsiot/microwave.py
pts211: Yeah it does! Not RGB though, just warm, natural, cool white. :)

abmantis (approved):
Thanks!

There is some stuff on the tests that should eventually be moved out of the test file and into a common place. But since all the tests need a refactoring to be updated to the new split api structure, and this is not yet going to the main branch, I think its fine for now to move forward and fix them all later.

PR #127, opened 2026-04-14
AWS IoT (2/4): Presence / online state

pts211:
## Summary

Wires up the receive side for `$aws/events/presence/{connected,disconnected}/{said}` MQTT topics. The subscriptions were already in `awsiot.Appliance.subscribe_topics()`; this PR routes the events through `_handle_mqtt_message` to a new `Appliance.update_online(bool)` method and has `get_online()` return the cached state instead of raising `NotImplementedError`.

Online state is `None` until the first presence event arrives.

## Stacking

Rebased onto `aws_iot` after #126 merged. This PR now contains only the presence / online-state changes.

## Test plan

- [x] `python -m pytest tests/test_microwave.py` - 14 passed, including 5 presence tests using the existing fake-MQTT fixture:
  - `get_online()` is `None` before any presence event
  - `connected` topic -> `True`
  - `disconnected` topic -> `False`
  - attr-callback fires on each transition
  - presence event for an unknown SAID is ignored
- [x] `python -m ruff check tests/test_microwave.py whirlpool/awsiot/appliance.py whirlpool/awsiot/appliancesmanager.py`
- [x] `basedpyright` clean on changed files

## Notes

Per @abmantis's review on #124: tests assert behavior at the wire boundary (MQTT topic + payload) rather than poking `_state` directly.

### DIFF
--- a/tests/test_microwave.py
+++ b/tests/test_microwave.py
@@ -340,3 +340,73 @@ async def test_cook_timer_time_complete_returns_timestamp(
         },
     )
     assert mwo.get_cook_timer_time_complete() == 1_776_101_159
+
+
+async def test_get_online_is_none_before_any_presence_event(
+    aws_manager: tuple[AwsAppliancesManager, FakeMqttClient],
+) -> None:
+    manager, _ = aws_manager
+    mwo = manager.microwaves[0]
+    assert mwo.get_online() is None
+
+
+async def test_presence_connected_sets_online_true(
+    aws_manager: tuple[AwsAppliancesManager, FakeMqttClient],
+) -> None:
+    manager, fake_mqtt = aws_manager
+    mwo = manager.microwaves[0]
+
+    fake_mqtt.inject(
+        f"$aws/events/presence/connected/{MWO_SAID}",
+        {"eventType": "connected", "clientId": "device", "timestamp": 1},
+    )
+    assert mwo.get_online() is True
+
+
+async def test_presence_disconnected_sets_online_false(
+    aws_manager: tuple[AwsAppliancesManager, FakeMqttClient],
+) -> None:
+    manager, fake_mqtt = aws_manager
+    mwo = manager.microwaves[0]
+
+    fake_mqtt.inject(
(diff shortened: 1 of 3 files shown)


tests/test_microwave.py
```diff
@@ -340,3 +340,73 @@ async def test_cook_timer_time_complete_returns_timestamp(
...
+    mwo = manager.microwaves[0]
+
+    fake_mqtt.inject(
+        f"$aws/events/presence/connected/{MWO_SAID}",
+        {"eventType": "connected", "clientId": "device", "timestamp": 1},
+    )
+    assert mwo.get_online() is True
+
+
+async def test_presence_disconnected_sets_online_false(
```
abmantis, line 366: we could use test parameterization and merge this test with the one above

tests/test_microwave.py
```diff
@@ -340,3 +340,73 @@ async def test_cook_timer_time_complete_returns_timestamp(
...
+        {"eventType": "connected", "clientId": "device", "timestamp": 1},
+    )
+    fake_mqtt.inject(
+        f"$aws/events/presence/disconnected/{MWO_SAID}",
+        {"eventType": "disconnected", "clientId": "device", "timestamp": 2},
+    )
+    assert mwo.get_online() is False
+
+
+async def test_presence_event_fires_attr_callback(
```
abmantis, line 383: we could also merge this one with the 2 above

whirlpool/awsiot/appliance.py
```diff
@@ -63,6 +64,12 @@ def update_state(self, new_state: dict[str, Any]):
         for callback in self._attr_changed:
             callback()
 
+    def update_online(self, online: bool):
+        """Update presence state and call callbacks."""
+        self._online = online
+        for callback in self._attr_changed:
```
abmantis, line 70: should we check if the online status changed before triggering the callbacks?

abmantis (approved):
LGTM! Thanks!
left 3 small suggestions which are not blockers so we can merge this.

PR #128, opened 2026-04-14
AWS IoT (3/4): Capability profile + gated microwave setters

### LINKED ISSUE #122: Implement capabilities
Implement capability checks to properly indicate what an appliance supports.

For this, we need to subscribe to the appliance capability topic, publish a message using the `CapabilityPartNumber` (from the appliance thing) as payload, and download the file from the link that is published on the subscribed topic.

```python
 self._mqtt.subscribe(
     f"api/capability/download/{self.appliance_info.model_number}/{self.appliance_info.said}/response",
 )
 self._mqtt.publish(
     f"api/capability/download/{self.appliance_info.model_number}/{self.appliance_info.said}",
     {"capabilityPartNumber": f"{thing_attrs.get('CapabilityPartNumber', '')}"},
 )
```

pts211:
Third PR in the AWS IoT microwave support chain.

Fixes #122

## Stacking

Rebased onto `abmantis:aws_iot` after #127 merged. This PR now contains the capability profile / gated setter work plus one #127 follow-up: duplicate presence events no longer fire attribute callbacks, with the presence tests parameterized per review.

Cook APIs (`start_cook`, `cancel_cook`, recipe execution) are deferred to the next PR (4/4).

## What this adds

- `whirlpool/awsiot/capabilities.py` - `CapabilityProfile` dataclass and `CapabilityDownloader` that fetches per-model capability files over the MQTT `api/capability/download/{model}/{said}` request/response topic pair (HTTPS body fetch included). Cached in memory by `capabilityPartNumber`, with 3-attempt exponential backoff retry.
- Capability-driven routing in `AwsAppliancesManager._add_appliance`: the `cooking` category now routes to `Microwave` only when the profile declares the `microwaveOven` feature, else to `Oven`. This closes the interim oven-regression introduced in #126.
- Six gated setters on `awsiot.Microwave`: `set_hood_fan_speed`, `set_hood_light_level`, `set_hood_light_color`, `set_control_locked`, `set_quiet_mode`, `set_sabbath_mode`. Each returns `False` + warns when the profile does not declare the addressee, and publishes the `{addressee, command: set, value}` MQTT command otherwise.
- Matching abstract methods on the `whirlpool.microwave.Microwave` ABC.
- Presence follow-up from #127: `update_online()` now skips callbacks when the online state is unchanged.

## Testing

- 13 parser/downloader unit tests in `tests/test_awsiot_capabilities.py` (two schemas, retry, cache, handle_message routing, HTTP-body fetch, inline-payload fallback).
- Microwave integration coverage in `tests/test_microwave.py` for profile exposure, microwave-vs-oven routing, skip-on-missing-CapabilityPartNumber, skip-on-download-failure, per-model cache sharing, all six setters in both supported and unsupported configurations, and the #127 presence callback follow-up.

## Local verification

- [x] `python -m pytest tests/test_awsiot_capabilities.py tests/test_microwave.py -q` - 37 passed
- [x] `python -m ruff check tests/test_awsiot_capabilities.py tests/test_microwave.py whirlpool/awsiot/capabilities.py whirlpool/awsiot/appliance.py whirlpool/awsiot/appliancesmanager.py whirlpool/awsiot/microwave.py whirlpool/microwave.py whirlpool/awsiot/aircon.py whirlpool/awsiot/dryer.py whirlpool/awsiot/oven.py whirlpool/awsiot/refrigerator.py whirlpool/awsiot/washer.py` - passed
- [x] `basedpyright` - 0 errors, 0 warnings, 0 notes

## Test plan

- [ ] CI passes against Python 3.12 / 3.13
- [ ] Maintainer review of the routing change (interim oven-regression fix)
- [ ] Real-device smoke test (capability fetch over MQTT -> correct routing + working hood/mode setters)


whirlpool/awsiot/capabilities.py
```diff
@@ -0,0 +1,224 @@
...
+        last_err: Exception | None = None
+        for attempt in range(self._retries):
+            try:
+                profile = await self._download(
+                    said, model_number, capability_part_number
+                )
+                self._cache[capability_part_number] = profile
+                return profile
+            except (TimeoutError, CapabilityDownloadError) as e:
+                last_err = e
```
abmantis, line 138: should we log something in `debug` here?

whirlpool/awsiot/capabilities.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+class CapabilityDownloader:
+    """Fetches and caches capability profiles per model (part number).
+
+    The MQTT client is expected to be already connected. The downloader does
+    not own topic subscriptions long-term — it subscribes/unsubscribes around
+    each request. Incoming responses are delivered by the AppliancesManager's
+    MQTT message dispatcher, which calls `handle_message` for every message;
+    the downloader consumes only topics matching pending requests.
```
abmantis:
```suggestion
    Issues an MQTT request for a capability file, awaits the response carrying
    a download URL, then fetches and parses the JSON body. Subscribes/unsubscribes
    around each request so subscriptions are not long-term.
    MQTT messages should be delivered to `handle_message()`.
    Results are cached per part number so repeated lookups are served without another
    round-trip.
```
A bit of rewording to describe a bit more what it does and avoid mentioning outside stuff like AppliancesManager.

whirlpool/awsiot/capabilities.py
```diff
@@ -0,0 +1,224 @@
...
+        raise CapabilityDownloadError(
+            f"Capability download for {said} failed after {self._retries} "
+            f"attempts: {last_err}"
+        )
+
+    async def _download(
+        self, said: str, model_number: str, capability_part_number: str
+    ) -> CapabilityProfile:
+        request_topic = f"api/capability/download/{model_number}/{said}"
+        response_topic = f"api/capability/download/{model_number}/{said}/response"
```
abmantis:
```suggestion
        request_topic = f"api/capability/download/{model_number}/{said}"
        response_topic = f"{request_topic}/response"
```

whirlpool/awsiot/capabilities.py
```diff
@@ -0,0 +1,224 @@
...
+                ) from e
+
+            raw = await self._fetch_body(response)
+            return parse_capability_profile(raw)
+        finally:
+            self._pending.pop(response_topic, None)
+            self._mqtt.unsubscribe(response_topic)
+
+    async def _fetch_body(self, response: dict[str, Any]) -> dict[str, Any]:
+        url = response.get("url") or response.get("downloadUrl")
```
abmantis: have you seen both `url` and `downloadUrl` for different appliances?

whirlpool/awsiot/capabilities.py
```diff
@@ -0,0 +1,224 @@
...
+        if isinstance(url, str) and url.startswith("http"):
```
abmantis:
```suggestion
        if not isinstance(url, str) or url.startswith("http"):
            return
```
lets invert it to reduce nesting

whirlpool/awsiot/capabilities.py
```diff
@@ -0,0 +1,224 @@
...
+                    f"Timed out waiting for capability response for {said}"
+                ) from e
+
+            raw = await self._fetch_body(response)
+            return parse_capability_profile(raw)
+        finally:
+            self._pending.pop(response_topic, None)
+            self._mqtt.unsubscribe(response_topic)
+
+    async def _fetch_body(self, response: dict[str, Any]) -> dict[str, Any]:
```
abmantis:
```suggestion
    async def _download_file(self, mqtt_response: dict[str, Any]) -> dict[str, Any]:
```
since this class mixes mqtt and http, I think this makes it a bit cleaer

whirlpool/awsiot/capabilities.py
```diff
@@ -0,0 +1,224 @@
...
+        retries: int | None = None,
+    ) -> None:
+        self._mqtt = mqtt_client
+        self._session = session
+        self._timeout = (
+            timeout if timeout is not None else CAPABILITY_DOWNLOAD_TIMEOUT
+        )
+        self._retries = (
+            retries if retries is not None else CAPABILITY_DOWNLOAD_RETRIES
+        )
```
abmantis:
```suggestion
```
we can remove timeout and retries and just use the constants directly

whirlpool/awsiot/capabilities.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def parse_capability_profile(raw: dict[str, Any]) -> CapabilityProfile:
+    """Normalize a raw capability JSON dict into a CapabilityProfile.
+
+    Supports two schemas:
+      1. Real-device files: top-level `partNumber`, `appliance.features` dict,
+         `cavities` dict for cavity addressees, top-level `hoodFan`/`hoodLight`
+         /`hoodLightColor` dicts.
+      2. Test fixtures: `capabilityPartNumber` string, `features` list,
```
abmantis: we should not have test-specific code. the test fixtures should be the same as a real file. isn't that possible?

whirlpool/microwave.py
```diff
@@ -88,20 +88,38 @@ def get_cook_timer_time_complete(self) -> int | None: ...
     @abstractmethod
     def get_hood_light_level(self) -> HoodLightLevel | None: ...
 
+    @abstractmethod
+    async def set_hood_light_level(self, level: HoodLightLevel) -> bool: ...
```
abmantis, line 92: maybe we can also add some `supports_*` methods, like `supports_hood_light_level() -> bool`, so that we can add entities conditionally on home assistant?

abmantis (changes requested):

whirlpool/awsiot/capabilities.py
pts211: Reworded as suggested — dropped the AppliancesManager reference.

whirlpool/awsiot/capabilities.py
pts211: Done — `response_topic = f"{request_topic}/response"`.

whirlpool/awsiot/capabilities.py
pts211: No — `downloadUrl` was speculative; it never appears in any captured response. Dropped it, only `url` now.

whirlpool/awsiot/capabilities.py
pts211: Inverted to an early return; the fetch is no longer nested.

whirlpool/awsiot/capabilities.py
pts211: Renamed to `_download_file(self, mqtt_response)`.

whirlpool/awsiot/capabilities.py
pts211: Added — each failed attempt now logs at debug with the attempt number and error.

whirlpool/awsiot/capabilities.py
pts211: Replaced the dual schema with a single real-schema parser; tests now load real capability files (`capability_mwo.json` / `capability_mwo_no_hood.json`) plus a few minimal real-schema inline dicts. Swapping to the real files surfaced two real bugs the synthetic fixtures were hiding — see the top-level comment.

whirlpool/awsiot/capabilities.py
pts211: Removed both params; `CapabilityDownloader` uses the module constants directly and the tests monkeypatch `CAPABILITY_DOWNLOAD_TIMEOUT` / `CAPABILITY_DOWNLOAD_RETRIES`.

whirlpool/microwave.py
pts211: Added `supports_hood_fan`/`supports_hood_light_level`/`supports_hood_light_color`/`supports_quiet_mode`/`supports_control_lock`/`supports_sabbath_mode` on the `Microwave` ABC, implemented over the real profile (hood sections, `quietMode` / `supportsHmiControlLockout` flags, derived sabbath). The setters now gate on these, so HA can add entities conditionally too.

pts211:
Heads up — moving the tests onto **real** captured capability files (instead of the synthetic test schema) surfaced two bugs the synthetic fixtures were masking, both now fixed in this push:

1. **Microwave was misrouted to `Oven`.** Routing used `profile.has_feature("microwaveOven")`, but in real capability files `microwaveOven` is a `cavities.*.cavityType`, not an `appliance.features` key (real microwave `features` is just `temperatureUnit`). So `has_feature` never matched and a real microwave constructed as an `Oven`. Routing now uses `has_cavity_type("microwaveOven")`.

2. **Quiet-mode / control-lock gating never matched.** Setters gated on `has_addressee("quietMode")` / `has_addressee("hmiControlLockout")`, but in real files those are top-level **bool flags** (`quietMode: true`, `supportsHmiControlLockout: false`), not addressees — so those setters were hard-blocked regardless of the model. Gating now reads the real flags, and the capability key (`supportsHmiControlLockout`) is decoupled from the MQTT command addressee (`hmiControlLockout`).

The `CapabilityProfile` is now generic primitives over the real schema (`features` / `cavity_types` / `sections` / `flags` / `sabbath_recipes_present`); the dead `addressees`/`commands`/`supports_command` model is gone. Verified against two real fixtures: W11788386 (with hood) and W11650001 (no hood).

whirlpool/awsiot/capabilities.py
```diff
@@ -0,0 +1,220 @@
...
+    Holds only generic primitives read directly from the real capability
+    file. Appliance classes layer their own semantic ``supports_*`` helpers
+    on top of these queries.
+    """
+
+    part_number: str
+    features: frozenset[str]
+    cavity_types: frozenset[str]
+    sections: frozenset[str]
+    flags: dict[str, bool] = field(default_factory=dict)
```
abmantis: can `flags` be a `frozenset[bool]` with only the `True` flags? or do you think we may need the `False` ones?

whirlpool/awsiot/capabilities.py
```diff
@@ -0,0 +1,220 @@
...
+    def has_feature(self, feature: str) -> bool:
+        return feature in self.features
+
+    def has_cavity_type(self, cavity_type: str) -> bool:
+        return cavity_type in self.cavity_types
+
+    def has_section(self, name: str) -> bool:
+        return name in self.sections
+
+    def flag(self, name: str, default: bool = False) -> bool:
```
abmantis:
```suggestion
    def has_flag(self, name: str) -> bool:
```
if we change to the frozenset above, lets also change here. we also don't need the default

whirlpool/awsiot/capabilities.py
```diff
@@ -0,0 +1,220 @@
...
+
+def parse_capability_profile(raw: dict[str, Any]) -> CapabilityProfile:
+    """Normalize a raw Whirlpool capability JSON dict into a CapabilityProfile.
+
+    The real schema is: top-level ``partNumber``; ``appliance.features`` dict;
+    ``cavities`` dict whose members may carry a ``cavityType`` and a
+    ``sabbathRecipes`` dict; top-level control sections such as ``hoodFan`` /
+    ``hoodLight`` / ``hoodLightColor``; and top-level boolean capability flags
+    such as ``quietMode`` and ``supportsHmiControlLockout``.
+    """
```
abmantis:
```suggestion
    """
```
This comment is a bit confusing. I think it is not needed since we have the json files in the repo

whirlpool/awsiot/capabilities.py
```diff
@@ -0,0 +1,220 @@
...
+        return cavity_type in self.cavity_types
+
+    def has_section(self, name: str) -> bool:
+        return name in self.sections
+
+    def flag(self, name: str, default: bool = False) -> bool:
+        return self.flags.get(name, default)
+
+
+def parse_capability_profile(raw: dict[str, Any]) -> CapabilityProfile:
```
abmantis: Think about this a bit more, maybe we should not generically add every section key blindly for two reasons:
- the reason is that the file is quite big and may have a lot of uninteresting sections
- we may want to know more than just section keys, or represent them in a different format. For example, in the future we may want to read the hoodFan enum speeds too, or have a list o recipes, etc.

Instead, what about having separate MicrowaveCapabilityProfile/AirconCapabilityProfile/etc, along with separate `parse_microwave_capability_profile`/`parse_aircon_capability_profile`/....?
That way we can tailor each class to the needs of each appliance. For example, an AirCon has no "cavity".

The new classes should only store things that are currently needed/used.
WDYT?

abmantis (changes requested):

abmantis:
@pts211 I've updated the code to address my last comments, as discussed on discord. I still want to take a pass at the tests, but it would be great if you can take a look at the production code already! Tks!

abmantis (approved):
Thanks a lot @pts211 !

I am still planing on refactoring the tests (the whole thing, not just this ones), so lets merge them as-is.

PR #130, opened 2026-04-15
AWS IoT: auto-reconnect MQTT client after unexpected disconnect

pts211:
## Summary

`MqttClient._on_disconnect` previously logged unexpected failures and cleared connection state, but never recovered. Paho's own reconnect path is not enough here because `connect()` installs a SigV4-signed websocket URL via `ws_set_options(path=...)`; reused signatures eventually expire.

This PR makes unexpected AWS IoT MQTT disconnects self-healing by rebuilding the paho client with fresh AWS credentials and a fresh SigV4-signed websocket URL.

## Evidence

Observed failure mode:

- One log line: `MQTT unexpected disconnect: Unspecified error`
- About 20 minutes later, AWS IoT entities went unavailable
- Hundreds of `Cannot publish, MQTT client not connected` and initial-state timeout warnings followed
- Recovery required restarting the host application

## Fix

On a failure disconnect, schedule a reconnect task that:

1. Stops the old paho client best-effort
2. Sleeps with exponential backoff (`1s -> 30s` cap)
3. Calls `self.connect()`, which fetches fresh AWS credentials when needed and creates a fresh SigV4-signed URL
4. Resubscribes existing `_subscribed_topics`

Additional reconnect hardening included here:

- Cached AWS credentials are invalidated when their `Expiration` timestamp is reached, so reconnects do not sign URLs with expired Cognito credentials.
- Paho auto-reconnect is disabled because it reuses the original signed websocket URL.
- `_connected` is cleared at the start of `connect()` to avoid stale event state.
- MQTT client IDs remain stable across reconnects, keeping appliance response topics valid.
- Callbacks from replaced/stale paho clients are ignored, so an old client cannot clear or disturb the active connection state.
- Explicit `disconnect()` cancels an in-flight reconnect task and tears down cleanly.

## Tests

`tests/awsiot/test_mqttclient_reconnect.py` covers:

- unexpected disconnect rebuilds the paho client and resubscribes
- clean disconnect does not reconnect
- reconnect keeps the MQTT client ID stable
- stale disconnect callbacks from old clients are ignored
- explicit disconnect cancels a pending reconnect

`tests/awsiot/test_auth.py` covers:

- expired AWS Cognito credentials are refreshed
- valid AWS Cognito credentials are reused from cache

## Local Verification

- [x] `python -m pytest tests/awsiot/test_auth.py tests/awsiot/test_mqttclient_reconnect.py tests/test_microwave.py -q` - 21 passed
- [x] `python -m ruff check whirlpool/awsiot/auth.py whirlpool/awsiot/mqttclient.py tests/awsiot/test_auth.py tests/awsiot/test_mqttclient_reconnect.py tests/test_microwave.py` - passed
- [x] `basedpyright whirlpool/awsiot/auth.py whirlpool/awsiot/mqttclient.py tests/awsiot/test_auth.py tests/awsiot/test_mqttclient_reconnect.py tests/test_microwave.py` - 0 errors, 0 warnings, 0 notes

### DIFF
--- a/whirlpool/awsiot/mqttclient.py
+++ b/whirlpool/awsiot/mqttclient.py
@@ -45,7 +45,7 @@ class MqttClient:
         self._reconnect_task: asyncio.Task[None] | None = None
         self._shutting_down: bool = False
 
-    async def connect(self) -> bool:
+    async def connect(self, *, log_failures: bool = True) -> bool:
         """Connect to the MQTT broker."""
         self._shutting_down = False
         self._connected.clear()
@@ -88,7 +88,10 @@ class MqttClient:
         try:
             client.connect(MQTT_ENDPOINT, port=443, keepalive=30)
         except Exception as e:
-            LOGGER.error("Failed to connect to MQTT broker: %s", e)
+            if log_failures:
+                LOGGER.error("Failed to connect to MQTT broker: %s", e)
+            else:
+                LOGGER.debug("Failed to connect to MQTT broker", exc_info=True)
             return False
 
         self._client = client
@@ -97,7 +100,10 @@ class MqttClient:
         try:
             await asyncio.wait_for(self._connected.wait(), timeout=10.0)
         except TimeoutError:
-            LOGGER.error("MQTT connection timeout")
+            if log_failures:
+                LOGGER.error("MQTT connection timeout")
+            else:
+                LOGGER.debug("MQTT connection timeout during reconnect")
             client.loop_stop()
             self._client = None
             return False
@@ -256,7 +262,7 @@ class MqttClient:
         """
         delay = RECONNECT_BACKOFF_INITIAL_SECONDS
         while not self._shutting_down:
-            LOGGER.warning("MQTT reconnecting in %.1fs", delay)
+            LOGGER.debug("MQTT reconnecting in %.1fs", delay)
             try:
                 await asyncio.sleep(delay)
             except asyncio.CancelledError:
@@ -277,17 +283,23 @@ class MqttClient:
                 except Exception:
                     LOGGER.debug("Error disconnecting old MQTT client", exc_info=True)
 
+            failure_logged = False
             try:
-                ok = await self.connect()
+                ok = await self.connect(log_failures=False)
             except asyncio.CancelledError:
                 return
-            except Exception:
-                LOGGER.exception("MQTT reconnect attempt raised")
+            except Exception as e:
+                LOGGER.warning("MQTT reconnect attempt failed: %s", e)
+                LOGGER.debug("MQTT reconnect attempt traceback", exc_info=True)
                 ok = False
+                failure_logged = True
 
             if ok:
-                LOGGER.warning("MQTT reconnected successfully")
+                LOGGER.info("MQTT reconnected successfully")
                 return
+
+            if not failure_logged:
+                LOGGER.warning("MQTT reconnect attempt failed")
 
             delay = min(delay * 2 if delay > 0 else 1.0, RECONNECT_BACKOFF_CAP_SECONDS)
 
--- a/whirlpool/awsiot/mqttclient.py
+++ b/whirlpool/awsiot/mqttclient.py
@@ -19,6 +19,7 @@ from .auth import Auth
 LOGGER = logging.getLogger(__name__)
 
 MQTT_ENDPOINT = "wt.applianceconnect.net"
+CONNECT_TIMEOUT_SECONDS = 10.0
 RECONNECT_BACKOFF_INITIAL_SECONDS = 1.0
 RECONNECT_BACKOFF_CAP_SECONDS = 30.0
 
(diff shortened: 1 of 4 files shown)


whirlpool/awsiot/mqttclient.py
```diff
@@ -60,6 +62,7 @@ async def connect(self) -> bool:
             protocol=MQTTProtocolVersion.MQTTv311,
             callback_api_version=CallbackAPIVersion.VERSION2,
         )
+        client._reconnect_on_failure = False  # type: ignore[attr-defined]  # noqa: SLF001
```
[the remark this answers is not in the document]
abmantis: This suggestion makes sense, right?

whirlpool/awsiot/mqttclient.py
```diff
@@ -42,11 +38,17 @@ def __init__(
...
         self._client_id: str | None = None
+        # Keep response topics stable across reconnects.
+        self._client_id_suffix: str = secrets.token_hex(8)
 
         self._loop = asyncio.get_running_loop()
+        self._reconnect_task: asyncio.Task[None] | None = None
+        self._shutting_down: bool = False
 
-    async def connect(self) -> bool:
+    async def connect(self, *, log_failures: bool = True) -> bool:
```
abmantis: I think its fine to always log, so we don't need this here

whirlpool/awsiot/mqttclient.py
```diff
@@ -85,16 +88,22 @@ async def connect(self) -> bool:
...
 
         try:
             await asyncio.wait_for(self._connected.wait(), timeout=10.0)
         except TimeoutError:
-            LOGGER.error("MQTT connection timeout")
+            if log_failures:
+                LOGGER.error("MQTT connection timeout")
+            else:
+                LOGGER.debug("MQTT connection timeout during reconnect")
             client.loop_stop()
```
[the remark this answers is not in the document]
abmantis: should we do this?

whirlpool/awsiot/mqttclient.py
```diff
@@ -194,27 +223,96 @@ def _on_message(
...
+                LOGGER.debug("MQTT reconnect attempt traceback", exc_info=True)
+                ok = False
+                failure_logged = True
+
+            if ok:
+                LOGGER.info("MQTT reconnected successfully")
+                return
+
+            if not failure_logged:
+                LOGGER.warning("MQTT reconnect attempt failed")
```
abmantis:
```suggestion
```
if we always log in `connect`, we can remove this

whirlpool/awsiot/mqttclient.py
```diff
@@ -194,27 +223,96 @@ def _on_message(
...
+                except Exception:
+                    LOGGER.debug("Error stopping old MQTT client loop", exc_info=True)
+                try:
+                    old.disconnect()
+                except Exception:
+                    LOGGER.debug("Error disconnecting old MQTT client", exc_info=True)
+
+            failure_logged = False
+            try:
+                ok = await self.connect(log_failures=False)
```
abmantis:
```suggestion
                if await self.connect(log_failures=False):
                    LOGGER.info("MQTT reconnected successfully")
                    return
```
so we can remove the `ok` var and the check from L297

whirlpool/awsiot/mqttclient.py
```diff
@@ -194,27 +223,96 @@ def _on_message(
...
+                failure_logged = True
+
+            if ok:
+                LOGGER.info("MQTT reconnected successfully")
+                return
+
+            if not failure_logged:
+                LOGGER.warning("MQTT reconnect attempt failed")
+
+            delay = min(delay * 2 if delay > 0 else 1.0, RECONNECT_BACKOFF_CAP_SECONDS)
```
abmantis: how can delay be non-positive?

abmantis (changes requested):

whirlpool/awsiot/mqttclient.py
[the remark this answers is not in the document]
pts211: Confirmed — paho 2.1.0's `Client` constructor accepts `reconnect_on_failure`. Switched to the constructor arg and dropped the private-attr access + `noqa: SLF001`.

whirlpool/awsiot/mqttclient.py
pts211: Removed the `log_failures` param. `connect()` now always logs failures at `debug` — AWS-side connect failure is non-fatal/expected and the caller surfaces the user-facing message.

whirlpool/awsiot/mqttclient.py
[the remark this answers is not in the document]
pts211: Done. The connect-timeout path (and `disconnect()`) now go through `_teardown_client()`, which does `loop_stop()` + `disconnect()`.

whirlpool/awsiot/mqttclient.py
pts211: Removed — `connect()` owns its failure logging now, so the reconnect loop no longer double-logs.

whirlpool/awsiot/mqttclient.py
pts211: Applied — dropped the `ok`/`failure_logged` vars; the loop is now `if await self.connect(): ...`.

whirlpool/awsiot/mqttclient.py
pts211: It can't — `delay` starts at 1.0 and only doubles, so the guard was dead. Removed it: `min(delay * 2, RECONNECT_BACKOFF_CAP_SECONDS)`.

abmantis (approved):
I think there is still an edge case that might fail: during a reconnect loop, if there is a disconnect right after connection, while `connect()` has not yet returned to `_reconnect_loop()`, it might never reconnect again (because the call to `_schedule_reconnect` will return since the task is already running).

But that should be hard to happen, so we can deal with that later.

Thanks!

PR #131, opened 2026-07-07
Add periodic keepalive fetch to AppliancesManager

### LINKED ISSUE #116: Possible cause for oven data timeouts

bdlcalvin:
Implements the keepalive discussed in #119 / #116, following your suggestion there: it lives in `AppliancesManager`, picks one appliance from `all_appliances`, and calls `fetch_data()` on it periodically. The task starts/stops with the event listener.

I went with 5 minutes since that matches the polling cadence Home Assistant has been using for washers/dryers for years (home-assistant/core#97222) without rate-limit issues — happy to change it if you'd rather match MizterB's field-proven 20 minutes.

Once this is released I'll follow up on the Home Assistant side: close home-assistant/core#175049 (integration-level polling, superseded by this), and the washer/dryer polling can then also be dropped from the integration as you mentioned.

Tests: added two for the keepalive; full suite passes locally. (Note: with `aioresponses==0.7.6` I had to pin `aiohttp<3.13` locally — 3.14 breaks aioresponses with a `ClientResponse.__init__` error. Unrelated to this change.)

Fixes #116

### DIFF
--- a/whirlpool/appliancesmanager.py
+++ b/whirlpool/appliancesmanager.py
@@ -1,5 +1,7 @@
+import asyncio
 import json
 import logging
+from contextlib import suppress
 from functools import cached_property
 from typing import Any
 
@@ -19,6 +21,11 @@ from .washer import Washer
 
 LOGGER = logging.getLogger(__name__)
 
+# The backend stops sending events over the websocket after a while unless
+# there is some REST activity on the account, so periodically fetch data for
+# one appliance to keep the event subscription alive.
+KEEPALIVE_INTERVAL_SECONDS = 5 * 60
+
 
 class AppliancesManager:
     def __init__(
@@ -31,6 +38,7 @@ class AppliancesManager:
         self._auth = auth
         self._session: aiohttp.ClientSession = session
         self._event_socket: EventSocket | None = None
+        self._keepalive_task: asyncio.Task[None] | None = None
         self._aircons: dict[str, Any] = {}
         self._dryers: dict[str, Any] = {}
         self._washers: dict[str, Any] = {}
@@ -86,8 +94,8 @@ class AppliancesManager:
             "ddm_cooking_bio_self_clean_tourmaline_v2",
             "ddm_cooking_bio_g3evo_pyro_bk_v1",
             "ddm_cooking_bio_self_clean_meat_probe_tourmaline_bk_v1",
-            "ddm_cooking_bio_self_clean_steam_tourmaline_v2",      # <-- W9 Ovens
-            "ddm_cooking_bi_mwo_self_clean_steam_tourmaline_v2",   # <-- W9 MWOs
+            "ddm_cooking_bio_self_clean_steam_tourmaline_v2",  # <-- W9 Ovens
+            "ddm_cooking_bi_mwo_self_clean_steam_tourmaline_v2",  # <-- W9 MWOs
         ]
 
         LOGGER.debug("Adding appliance %s", appliance_data)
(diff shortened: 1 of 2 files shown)


whirlpool/appliancesmanager.py
```diff
@@ -204,14 +212,37 @@ async def start_event_listener(self):
...
+        """Periodically fetch data for one appliance to keep events flowing."""
+        while True:
+            await asyncio.sleep(KEEPALIVE_INTERVAL_SECONDS)
+            appliance = next(iter(self.all_appliances.values()), None)
+            if appliance is None:
+                continue
+            try:
+                await appliance.fetch_data()
+            except (aiohttp.ClientError, TimeoutError) as ex:
+                LOGGER.warning(f"Keepalive fetch failed: {ex}")
```
abmantis:
```suggestion
                LOGGER.warning("Keepalive fetch failed: %s", ex)
```

whirlpool/appliancesmanager.py
```diff
@@ -204,14 +212,37 @@ async def start_event_listener(self):
...
 
+    async def _keepalive(self):
+        """Periodically fetch data for one appliance to keep events flowing."""
+        while True:
+            await asyncio.sleep(KEEPALIVE_INTERVAL_SECONDS)
+            appliance = next(iter(self.all_appliances.values()), None)
+            if appliance is None:
+                continue
+            try:
+                await appliance.fetch_data()
```
abmantis, line 242: lets catch a broader exception, so that the task does not die silently if something unexpected happens

abmantis (changes requested):

bdlcalvin:
The failing pytest jobs here are pre-existing: aioresponses breaks on aiohttp 3.14, which CI now installs since the dev requirements leave it unbounded — #132 fixes that. With aiohttp pinned locally, the full suite (incl. the new keepalive tests) passes.

abmantis:
You can merge `master` to fix the aiohttp test issue.

whirlpool/appliancesmanager.py
bdlcalvin: Done in 9270e67.

whirlpool/appliancesmanager.py
bdlcalvin: Done in 9270e67 — catching `Exception` now so the task never dies silently.

bdlcalvin:
Merged master and ported the keepalive tests to aiointercept — full suite passes locally on aiohttp 3.14 now. Both review comments addressed in 9270e67.

abmantis (approved):
Thanks!

abmantis:
[v1.3.0](https://pypi.org/project/whirlpool-sixth-sense/1.3.0/) is now released with this change.

PR #137, opened 2026-07-14
fix: return int from KitchenTimer get_remaining_time/get_total_time

matthewdias:
`KitchenTimer.get_remaining_time()` and `get_total_time()` called `_get_attribute()`, which returns `str | None`. Consumers doing arithmetic on the result (e.g. `timedelta(seconds=timer.get_remaining_time())`) raise `TypeError` against a real device because the value is a numeric string.

This switches both methods to `_get_int_attribute()`, matching the washer/dryer `get_time_remaining()` counterparts, and adds `int | None` return annotations.

Tests: gave SAIDOVEN1's kitchen-timer fixture attributes non-zero values (`SetTimeSet=300`, `StatusTimeRemaining=180`, `StatusState=1`) and added assertions for the int-typed returns plus `get_state()` for both a running (oven1) and standby (oven2) timer. The new assertions fail without the fix (`assert '300' == 300`) and the full suite passes with it.

Context: found while adding a kitchen-timer sensor to the Home Assistant whirlpool integration (home-assistant/core#176463).

### DIFF
--- a/tests/data/mock_data.json
+++ b/tests/data/mock_data.json
@@ -1030,7 +1030,7 @@
         "updateTime": 1642158986086
       },
       "KitchenTimer01_StatusTimeRemaining": {
-        "value": "0",
+        "value": "180",
         "updateTime": 1642158986082
       },
       "Sys_AlertStatusCustomerFaultCodeNotification": {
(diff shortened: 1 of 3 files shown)


abmantis (approved):
Thanks!

Issue #138, opened 2026-07-14
Fixtures: Maytag MFW7020RF0 washer + MGD7020RF0 dryer (US) — dryer detected as `type=washer`


arbog:
Capturing washer+dryer fixtures to help with new-API support, per your request in home-assistant/core#151547. Used `tools/capture_fixtures.py` on the `aws_iot` branch (`710a875`) with `--redact`, Python 3.14.

## Setup
- Brand **Maytag**, region **US**
- Washer `MFW7020RF0` (systemVersion 2.1.0), Dryer `MGD7020RF0` (systemVersion 3.0.0)

## Discovery — dryer mis-classified
```
Discovered 2 appliance(s):
  <washer SAID redacted>  washer   washer
  <dryer SAID redacted>   washer   Dryer
```
Both come back as `type=washer`. The **dryer (`MGD7020RF0`) is classified as a washer**, though its state payload is clearly a dryer (`dryLevel`, `dryTemperature`, `wrinkleShield`, `staticGuardEnable`, `dampDry`, `lintTrap`, `drumLight`, `lowAirFlow`). Likely why HA can't set the dryer up.

## Redacted fixtures
**`thing_washer_MFW7020RF0-a549e9cc.json`**
```json
{
  "thingName": "a549e9cc",
  "thingTypeName": "MFW7020RF0",
  "attributes": {
    "Name": "776173686572",
    "Category": "Laundry",
    "Serial": "REDACTED"
  }
}

```

**`state_washer_MFW7020RF0-a549e9cc_full.json`**
```json
{
  "washer": {
    "applianceState": "standby",
    "cycleName": "normal",
    "cycleType": "standard",
    "specialName": "",
    "currentPhase": "",
    "soilLevel": "extraHeavy",
    "washTemperature": "warm",
    "spinSpeed": "medium",
    "extraRinse": "off",
    "postCare": "off",
    "extraPower": "off",
    "pets": "off",
    "steam": "off",
    "cycleTime": {
      "state": "idle",
      "time": 5351,
      "timeComplete": 1783898525,
      "timePaused": 1782754709
    },
    "delayTime": {
      "state": "idle",
      "time": 0,
      "timeComplete": 0,
      "timePaused": 0
    },
    "sessionId": "",
    "cleanWasher": false,
    "doorStatus": "closed",
    "doorLockStatus": false
  },
  "remoteStartEnable": false,
  "hmiControlLockout": false,
  "faultHistory": [
[...]

PR #139, opened 2026-07-15
fix(awsiot): detect dryers that report category "laundry" instead of "fabriccare"

arbog:
## Problem

On `aws_iot`, `AppliancesManager._add_appliance` routes `category == "laundry"` to `Washer` and expects dryers to report `fabriccare`. But some dryers report `Category: "Laundry"` (same as washers), so they get classified as washers and no `Dryer` is created.

This is the washer + dryer case from #138 (and home-assistant/core#151547): a Maytag `MGD7020RF0` dryer is discovered as `type=washer`.

## Fix

Disambiguate the `laundry` category by model number. Whirlpool/Maytag/KitchenAid laundry model numbers encode the product type in the third character (`D` = dryer, `W` = washer), so a laundry appliance whose model looks like a dryer is routed to `Dryer`. Washer behavior is unchanged.

## Evidence (redacted fixtures, per #138)

| appliance | model | `Category` | state root key |
|-----------|-------|------------|----------------|
| washer | `MFW7020RF0` | `Laundry` | `washer` |
| dryer  | `MGD7020RF0` | `Laundry` | `dryer` (was mis-typed as washer) |

## Open question

The third-character heuristic matches the models I can see (US Maytag), but you know the model scheme far better. The MQTT state also has a definitive `washer`/`dryer` root key, though it's only available after `fetch_data()`. If you'd prefer a capability- or state-based signal, I'm happy to rework — deferring to your preference.

## Testing

`ruff check` and a parametrized test for the model → type helper pass locally.

---
*Disclosure: this change was drafted with AI assistance (Claude) and reviewed by me before submitting, per this repo's `AGENTS.md` and common AI-disclosure practice.*

### DIFF
--- a/tests/awsiot/test_appliancesmanager.py
+++ b/tests/awsiot/test_appliancesmanager.py
@@ -1,0 +1,20 @@
+import pytest
+
+from whirlpool.awsiot.appliancesmanager import _is_dryer_model
+
+
+@pytest.mark.parametrize(
+    ("model_number", "expected"),
+    [
+        ("MGD7020RF0", True),  # Maytag gas dryer
+        ("MED7020RF0", True),  # Maytag electric dryer
+        ("WGD5620HW1", True),  # Whirlpool gas dryer
+        ("MFW7020RF0", False),  # Maytag front-load washer
+        ("MHW6630HW0", False),  # Maytag front-load washer
+        ("WTW5010LW0", False),  # Whirlpool top-load washer
+        ("", False),
+        ("MF", False),
+    ],
+)
+def test_is_dryer_model(model_number: str, expected: bool) -> None:
+    assert _is_dryer_model(model_number) is expected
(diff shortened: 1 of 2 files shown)


arbog:
FYI: the failing `pytest` job is pre-existing on `aws_iot` (the branch's own CI has been red on recent pushes), not from this change — the diff only touches the `laundry` classification, and the added `test_is_dryer_model` cases pass (`ruff` and `pyright` are green too). Happy to rebase once the suite's sorted.

abmantis (commented):
Hey! Thanks for opening this PR. 
I wonder if we could do this using capabilities instead? Would it be cleaner?
[Support for capabilities](https://github.com/abmantis/whirlpool-sixth-sense/pull/128) is almost done, just need to do a final review on it.

arbog:
makes sense — capabilities would be cleaner and more robust than the model-string heuristic (that's the "open question" i flagged). happy to rework this on top of #128 once it lands, or close this in favor of you folding laundry disambiguation into the capability work — whichever you prefer. thanks for the quick look!

abmantis:
> in favor of you folding laundry

I really dislike folding laundry :shirt: :rofl: 

Since #128 is merged, are you OK with redoing this around capabilities?

PR #155, opened 2026-08-10
Add KitchenAid oven model

Goldenhawk2003:
this is my model Kitchenaid oven, it is not in the list. I need to add it so that I can use the library pls accept this change


abmantis (approved):
Thanks!

PR #161, opened 2026-08-25
fix(awsiot): capability downloadUrl, nested state merge, cook timer "run" command, paho-mqtt dep

pts211:
Found while live-testing #160 against a KitchenAid KMML550RPS00 microwave. None of these are in the CLI PR itself; they're all in `aws_iot` library code, but the first one blocks the CLI from listing anything at all and the third means `start_cook` never honoured its duration.

### 1. Capability download parses the wrong key (blocker)
The broker's reply on `api/capability/download/<model>/<said>/response` is
```json
{"responseCode": 200, "downloadUrl": "https://iot-evolution-cooking-category.s3.amazonaws.com/PROD/Appliance/<part>?X-Amz-..."}
```
`CapabilityDownloader._download_file` looked for `"url"`, so the MQTT envelope itself was handed to `_read_part_number` and every appliance failed with `Capability file missing 'partNumber' — skipping`. Fixed to read `downloadUrl` (with `url` kept as a fallback); tests updated to use the real response shape.

### 2. `Appliance.update_state` replaced the dict with partial deltas
`dt/<model>/<said>/state/update` only carries what changed, and the deltas are nested partials. Captured while a cook started:
```json
{"primaryCavity": {"mwoPowerLevel": 30}}
{"primaryCavity": {"recipeId": "microwave"}}
{"primaryCavity": {"cavityState": "cooking", "turnTable": "on", "cookTimer": {"time": 20}}}
{"primaryCavity": {"recipeExecutionState": "RUNNING"}}
{"primaryCavity": {"cavityLight": true, "cookTimer": {"state": "running", "timeComplete": 1787702389}, "sessionId": "..."}}
```
Replacing the dict (or even shallow-merging it) dropped door/recipe/fan/etc. to `None` after any setter until the next `getState`. Now merges recursively. Added `tests/awsiot/test_appliance_state.py`.

### 3. `start_cook` sent `cookTimer.command: "start"` — duration was ignored
Every cook ran for exactly 5 s regardless of `duration_seconds` (the recipe's capability-file default: `cookTime.range.default = 5`). The appliance's timer commands are **`run` / `set` / `add`** — there is no `start` — and an unknown command is silently ignored (the `run` itself is still `accepted`). From the KitchenAid app decompile: `RecipeCommand.getCommandJSONString` serialises `cookTimer: {command, time}` and the start-cycle path picks `command = "run"` (`vk.b` enum → `Constants.Commands.COMMAND_RUN`; `set` is used to modify a running timer, `add` for add-time). Verified live: requested 20 s → `cookTimer.time: 20`, `running` → `completed` in 20.0 s. Tried and rejected along the way: top-level `cookTime` (rejected `C000`), `time` as float (still 5 s).

### 4. `set_cook` validation didn't match the capability file (review follow-up)
`power_level` was checked 1..100 and `duration_seconds` >= 1, but the recipe's `whrOptions` say otherwise (`microwave.mwoPowerLevel` 10100 step 10; `cookTime` min 5; `keepWarm` 9003600 step 900), and only `microwave` has an editable power level  `reheat`/`defrost`/`soften` fix it under `nonEditableOptions`. The profile now carries `recipes[name] -> MicrowaveRecipeOptions` and `set_cook(recipe, duration_seconds, power_level=None)` validates against it, sending `mwoPowerLevel` only when editable. Command acks/rejections on `cmd/.../response` are no longer merged into state.

### 5. `paho-mqtt` missing from `pyproject.toml`
It was only in `requirements.txt`, so `pip install -e .` followed by `import whirlpool` raised `ModuleNotFoundError: paho`.

### Verified live
With these applied, `cli.py -l` lists the microwave and the #160 menu works end to end: status, cavity light on/off, hood fan Low/Off, hood light Low/Off, raw dump, and `t` (Microwave recipe, power 30, 20 s) running for the full 20 s then `o` cancel. State stayed coherent through every push. `pytest`: 154 passed; `ruff check` and `basedpyright` clean.

Side observations, not addressed here: `cookTimer.timeComplete` is an epoch timestamp (the CLI prints it as "time complete (seconds)"); `ovenDisplayTemperature: 25` with `temperatureUnit: fahrenheit` looks like °C; after a cook completes the recipe sits in step 1 (the "addTime" user prompt) with `recipeExecutionState: RUNNING` until cancelled, and a new `run` while there is rejected `C103`; the app sends `turntable` (lowercase) as a boolean in commands while state reports `turnTable: "on"/"off"`.

Note: `ruff format --check` reports pre-existing drift in 9 files unrelated to this change (newer ruff than the pinned hook) — left untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AK2q3npAU3S4xhta4QD8YN

### DIFF
--- a/whirlpool/awsiot/appliance.py
+++ b/whirlpool/awsiot/appliance.py
@@ -28,6 +28,18 @@ def gated_set(support_check, label):
     return decorator
 
 
+def _deep_merge(base: dict[str, Any], update: dict[str, Any]) -> dict[str, Any]:
+    """Return a copy of `base` with `update` merged in, recursing into dicts."""
+    merged = dict(base)
+    for key, value in update.items():
+        existing = merged.get(key)
+        if isinstance(existing, dict) and isinstance(value, dict):
+            merged[key] = _deep_merge(existing, value)
+        else:
+            merged[key] = value
+    return merged
+
+
 class Appliance(BaseAppliance):
     """Whirlpool awsiot appliance class"""
 
@@ -72,7 +84,7 @@ class Appliance(BaseAppliance):
         so replacing the dict would drop everything else until the next
         getState round-trip.
         """
-        self._data_dict = {**self._data_dict, **new_state}
+        self._data_dict = _deep_merge(self._data_dict, new_state)
         self._initial_data_event.set()
         for callback in self._attr_changed:
             callback()
--- a/whirlpool/awsiot/appliance.py
+++ b/whirlpool/awsiot/appliance.py
@@ -66,8 +66,13 @@ class Appliance(BaseAppliance):
         )
 
     def update_state(self, new_state: dict[str, Any]):
-        """Update appliance state and call callbacks."""
-        self._data_dict = new_state
+        """Merge a (possibly partial) state update and call callbacks.
+
+        `dt/.../state/update` messages only carry the attributes that changed,
+        so replacing the dict would drop everything else until the next
+        getState round-trip.
+        """
+        self._data_dict = {**self._data_dict, **new_state}
         self._initial_data_event.set()
         for callback in self._attr_changed:
             callback()
--- a/whirlpool/awsiot/capabilities.py
+++ b/whirlpool/awsiot/capabilities.py
@@ -15,7 +15,7 @@ import asyncio
 import json
 import logging
 import uuid
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Any
 
 import aiohttp
@@ -31,6 +31,35 @@ class CapabilityDownloadError(Exception):
 
 
 @dataclass(frozen=True)
+class OptionRange:
+    """Inclusive numeric range with a step, as declared under whrOptions."""
+
+    min: int
+    max: int
+    step: int
+    default: int
+
+    def contains(self, value: int) -> bool:
+        if not self.min <= value <= self.max:
+            return False
+        return self.step <= 1 or (value - self.min) % self.step == 0
+
+
+@dataclass(frozen=True)
+class MicrowaveRecipeOptions:
+    """Editable options of one cavity recipe (e.g. "microwave", "reheat").
+
+    `power_level` is the editable mwoPowerLevel range, or None when the recipe
+    has a fixed power level (listed under nonEditableOptions) that must not be
+    sent. `cook_time` is the editable cookTime range in seconds.
+    """
+
+    cook_time: OptionRange
+    power_level: OptionRange | None = None
+    fixed_power_level: int | None = None
+
+
+@dataclass(frozen=True)
 class MicrowaveCapabilityProfile:
     """The capability switches the Microwave class currently uses."""
 
@@ -41,6 +70,7 @@ class MicrowaveCapabilityProfile:
     supports_quiet_mode: bool
     supports_control_lock: bool
     supports_sabbath_mode: bool
+    recipes: dict[str, MicrowaveRecipeOptions] = field(default_factory=dict)
 
 
 def _read_part_number(raw: dict[str, Any]) -> str:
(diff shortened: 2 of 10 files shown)


whirlpool/awsiot/capabilities.py
```diff
@@ -187,7 +187,8 @@ async def _download(
             self._mqtt.unsubscribe(response_topic)
 
     async def _download_file(self, mqtt_response: dict[str, Any]) -> dict[str, Any]:
-        url = mqtt_response.get("url")
+        # The broker replies with {"responseCode": 200, "downloadUrl": "..."}.
+        url = mqtt_response.get("downloadUrl") or mqtt_response.get("url")
```
abmantis: is `url` ever sent?

whirlpool/awsiot/microwave.py
```diff
@@ -263,7 +263,10 @@ async def set_cook(
                 "addressee": "primaryCavity",
                 "recipeID": recipe.value,
                 "mwoPowerLevel": float(power_level),
-                "cookTimer": {"command": "start", "time": duration_seconds},
+                # Timer commands are run/set/add (no "start"); the app uses
+                # "run" when starting a recipe. An unknown command is ignored
+                # and the recipe falls back to its default cookTime (5 s).
```
abmantis:
```suggestion
```
We can leave this out. No need to mention "start" if it was wrong before

whirlpool/awsiot/capabilities.py
```diff
@@ -187,7 +187,8 @@ async def _download(
             self._mqtt.unsubscribe(response_topic)
 
     async def _download_file(self, mqtt_response: dict[str, Any]) -> dict[str, Any]:
```
abmantis, line 279: should we also check that the response code is 200, before calling `_download_file()`?

whirlpool/awsiot/appliance.py
```diff
@@ -66,8 +78,13 @@ async def subscribe_topics(self):
...
     def update_state(self, new_state: dict[str, Any]):
-        """Update appliance state and call callbacks."""
-        self._data_dict = new_state
+        """Merge a (possibly partial) state update and call callbacks.
+
+        `dt/.../state/update` messages only carry the attributes that changed,
+        so replacing the dict would drop everything else until the next
+        getState round-trip.
+        """
+        self._data_dict = _deep_merge(self._data_dict, new_state)
```
abmantis, line 87: this will also merge command acks fields like `responseCode` and `addressee`, right? Not sure if we shoul do that tbh

abmantis (changes requested):
tests/data/awsiot/capability_mwo.json shows two more instances of that same failure mode, still unguarded at microwave.py:251-252:

- power_level is validated 1..100, but microwave's mwoPowerLevel range is {min:10, max:100, step:10, default:100}. set_cook(Recipe.Microwave, 55, 60) passes validation and cooks at 100%, returning True. cli_microwave_menu.py:131 prompts "Enter power level (1-100)".
- duration_seconds is validated >= 1, but cookTime is {min:5, ...} (and keepWarm is {min:900, max:3600, step:900}). Sub-minimum durations fall back to the recipe default — the identical bug this PR just fixed.
- set_cook always sends mwoPowerLevel, but only microwave lists it under requiredOptions; reheat/defrost/soften/melt/mwCook have it under nonEditableOptions. Only Recipe.Microwave was verified live, and test_start_cook_publishes_expected_payload only covers that one.

whirlpool/awsiot/capabilities.py
pts211: No — only `downloadUrl` is ever sent (the `url` key was an assumption in #128). Dropped the fallback in ff53145.

whirlpool/awsiot/microwave.py
pts211: Done, comment removed in ff53145.

whirlpool/awsiot/capabilities.py
pts211: Yes, good call — `get()` now raises `CapabilityDownloadError` when the response carries a non-200 `responseCode` before touching `_download_file()`, with a test (`test_non_200_response_code_is_download_error`). ff53145.

whirlpool/awsiot/appliance.py
pts211: Right, it would have. Fixed one level up in `_handle_mqtt_message` (ff53145): a `cmd/.../response` is only treated as state when `response == "accepted"` **and** it carries a non-empty `payload` (i.e. the getState reply). Plain acks (`{"response":"accepted"}`) and rejections (`{"payload":{"errorCode":"C000"}}`) are ignored. Test: `test_command_acks_and_rejections_do_not_touch_state`. `responseCode` only appears on the capability topic, which is consumed earlier by the downloader.

pts211:
Thanks — all addressed in ff53145 (154 tests, ruff + basedpyright clean):

**Validation now comes from the capability file.** `parse_microwave_capability_profile` collects the cavity's timed recipes into `profile.recipes[name] = MicrowaveRecipeOptions(cook_time=OptionRange, power_level=OptionRange | None, fixed_power_level)`. `set_cook` uses it:

- `duration_seconds` must satisfy the recipe's `cookTime` range **and step** (microwave/reheat/…: 5–5940 s; `keepWarm`: 900–3600 step 900). Sub-minimum durations raise instead of silently running the default.
- `mwoPowerLevel` is only sent when the recipe lists it under `requiredOptions` (only `microwave`: 10–100 step 10 — so 55 now raises instead of cooking at 100%). For `reheat`/`defrost`/`soften`/`melt`/`mwCook` (fixed power under `nonEditableOptions`) it's omitted, and passing one raises `ValueError("recipe 'reheat' has a fixed power level of 70; omit power_level")`.
- Unknown recipe for this model → `ValueError`.

That meant an API change: `set_cook(recipe, duration_seconds, power_level=None)` (was `(recipe, power_level, duration_seconds)`). The #160 menu calls it with keywords so it still runs, but it always prompts for power, which will now raise for the fixed-power recipes — suggest prompting only when `mw.get_recipe_options(recipe).power_level is not None` (the range's min/max/step is right there for the prompt text too). Happy to do that on #160 if you'd like.

Also per your inline comments: `responseCode` checked before download, `url` fallback dropped, command acks/rejections no longer merged into state, and the test mock now answers `getState` in the broker's real `{requestId, response, payload}` shape (it was missing `response`, which is how the ack bug slipped past the tests).

Live: `microwave` recipe re-verified with the new code path (same payload as before). The fixed-power recipes omitting `mwoPowerLevel` are covered by tests against the real capability file but I haven't run a reheat on the appliance yet — can do if you want it confirmed.

abmantis (approved):
Thanks!

PR #166, opened 2026-09-01
fix(awsiot): parenthesize multi-exception except clauses

arbog:
## What

The AWS IoT backend shipped in #162 contained two Python 2-style `except A, B:` clauses (in `whirlpool/awsiot/appliancesmanager.py` and `whirlpool/awsiot/capabilities.py`), which are a `SyntaxError` under Python 3 and make the entire `whirlpool.awsiot` module unimportable.

## Why

Any import of the newer backend fails at parse time — washer, dryer, and microwave code alike.

## Change

- `except (ValueError, UnicodeDecodeError):`
- `except (KeyError, TypeError, ValueError):`

## Note

`ruff format` (0.16.5) currently wants to strip these parentheses and regenerate the invalid form, so `ruff format --check` will flag these two lines. Recommend pinning/upgrading ruff before merging.

### DIFF
--- a/whirlpool/awsiot/capabilities.py
+++ b/whirlpool/awsiot/capabilities.py
@@ -110,7 +110,7 @@ def _option_range(option: Any) -> OptionRange | None:
             step=int(rng.get("step", 1)),
             default=int(rng.get("default", rng["min"])),
         )
-    except KeyError, TypeError, ValueError:
+    except (KeyError, TypeError, ValueError):
         return None
 
 


arbog:
Closing this — it was based on a wrong assumption on my part.

`except ValueError, UnicodeDecodeError:` (bare exception tuple) is a SyntaxError on Python ≤3.13, but this project declares `requires-python = ">=3.14"`, and Python 3.14 accepts bare exception tuples. The upstream code is correct as-is, and `ruff format` (targeting py314) intentionally prefers the unparenthesized form — it is not a formatter bug.

Apologies for the noise.

PR #172, opened 2026-09-06
Move MQTT client to worker thread

abmantis:


pts211:
Tested `67a1dcc` in HA 2026.7.4 with a KitchenAid KMML550RPS00. Startup, state updates, hood-light control, and integration unload/reload all worked.

Forced MQTT connection loss recovered in 1.30s, then 1.24s after reload. Client ID and subscriptions stayed stable, fresh state reads worked afterward, and callbacks ran on HA’s main thread. No blocking-call or thread-safety warnings during those tests.

Two small observations:
- `test_failed_reconnect_logs_concise_warning_and_debug_traceback` failed once because the debug traceback hadn’t been captured yet. It may need to wait for that record before checking the logs. All 153 tests passed on rerun.
- Confirmed the disconnect-during-connect TODO: with no CONNACK, disconnect waited about 10 seconds before completing.

Looks good from my testing 👍

